PR #1, opened 2019-08-19
Character creation branch

willadams05:
Added a character creation (CC) scene which allows users to select their starting class.


EthanVieira (approved):
Good

PR #7, opened 2019-09-29
Lumbridge branch

CurtisGreen:
Added world map, added cookies to save data, and merged in menu changes


source/scenes/menu.js
```diff
@@ -19,44 +20,74 @@ export class MenuScene extends Phaser.Scene{
...
+            // Let level know we have cookies
+            this.levelData.hasCookies = true;
+
+            // Remove starting whitespace
+            while (cookie[0] == ' ') {
+                cookie = cookie.substring(1);
+            }
+
+            // Check for all stored cookies
+            for (let k = 0; k < cookieArray.length; k++) {
```
EthanVieira: find a way to loop through and set these easily and get rid of the for loop you already have

EthanVieira (changes requested):

source/scenes/menu.js
CurtisGreen: Okay

PR #8, opened 2019-10-01
Character data cleanup

CurtisGreen:
1) Cookies are now stored in an object called "characterData" that's easier to pass around (this.characterData = characterData instead of this.gold = gold; this.class = class; etc.)

2) Cookies are now stored as a JSON string which simplifies the get/set cookies functions

3) Cookies now store autoclickers (in a rudimentary way, will update later when more purchasing options are available)

4) Cookies now store level completion

### DIFF
--- a/source/level.js
+++ b/source/level.js
@@ -19,8 +19,10 @@ export class Level extends Phaser.Scene{
         damageByClicking: 0,
         damageByAutoClick: 0,
         numberOfAutoClickers: 0,
-        levelCompletions: {
-        	lumbridge: false
+        unlockedLevels: {
+        	tutorialIsland: true,
+        	lumbridge: true,
+        	varrock: false
         }
     };
     // Autoclickers
@@ -149,7 +151,7 @@ export class Level extends Phaser.Scene{
         this.questCompleteText = this.add.text(530, 290, 'Quest complete!', {fill: 'white'}).setDepth(3);
 
         // If level quest has already been completed
-        if (!this.characterData.levelCompletions.lumbridge) {
+        if (!this.characterData.unlockedLevels.lumbridge) {
         	this.questCompleteText.visible = false;
         }
 
@@ -223,7 +225,7 @@ export class Level extends Phaser.Scene{
                 // Store for cookies
                 switch(this.background.name) {
                 	case "lumbridge":
-                		this.characterData.levelCompletions.lumbridge = true;
+                		this.characterData.unlockedLevels.varrock = true;
                 		break;
                 }
                 console.log("Quest complete!");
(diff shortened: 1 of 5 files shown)


source/level.js
```diff
@@ -10,22 +10,27 @@ export class Level extends Phaser.Scene{
...
-	characterClass = '';
+    characterData = {
+        gold: 0,
+        characterClass: '',
+        enemiesKilled: 0,
+        timesClicked: 0,
+        damageByClicking: 0,
+        damageByAutoClick: 0,
+        numberOfAutoClickers: 0,
+        levelCompletions: {
```
EthanVieira: I don't think we will consider levels to be "completed" but more so levels will be unlocked after doing certain things/gaining certain levels. So maybe change this to levelUnlocks or something (true meaning it is unlocked).

source/level.js
```diff
@@ -173,49 +201,47 @@ export class Level extends Phaser.Scene{
...
 
             // Check quest completion
-            if (this.enemiesKilled == this.killQuest){
+            if (this.characterData.enemiesKilled == this.killQuest){
                 this.questCompleteText.visible = true;
+
+                // Store for cookies
+                switch(this.background.name) {
+                	case "lumbridge":
+                		this.characterData.levelCompletions.lumbridge = true;
```
EthanVieira: Same thing here

source/level.js
```diff
@@ -138,23 +138,51 @@ export class Level extends Phaser.Scene{
...
+        for (let i = 0; i < this.autoClickers.length; i++) {
+            this.autoClickers[i].pause = false;
+        }
+       
+        // Re-add autoclickers from cookies on first load
+        if (this.characterData.hasCookies && this.autoClickers.length == 0){
+            for (let i = 0; i < this.characterData.numberOfAutoClickers; i++) {
+                let dps = 5;
+                let level = 1;
+                let type = 'Hired Bowman';
```
EthanVieira, line 176: nice. this gives me ideas for autoclickers like "join a clan" - "get a gf"

CurtisGreen: Gotta carve and forge that diamond ring on your own to show you truly care

source/level.js
CurtisGreen: K, with this change when you complete the level quest it unlocks the next level rather than your own. For example tutorial island unlocks lumbridge, lumbridge unlocks varrock etc.
For now I'm going to set tutorial island and lumbridge to unlocked until tutorial island is created

source/level.js
CurtisGreen: See above and most recent commit

PR #10, opened 2019-10-19
Multiple enemies

CurtisGreen:
Levels now support multiple different enemies, cycles through them randomly. Added goblin to Lumbridge. Also added line breaks between functions so it's not as crammed.

### DIFF
--- a/source/enemy.js
+++ b/source/enemy.js
@@ -14,8 +14,12 @@ export class Enemy {
         this.enemy = data.scene.add.image(data.x, data.y, data.name);
         this.enemy.setOrigin(.5,0).setDepth(2).setScale(.4);
         this.enemy.setInteractive();
+        this.enemy.visible = false;
         this.enemy.on("pointerup", ()=>{
-            this.clickEnemy();
+            // Check if it is the current enemy
+            if (this.enemy.visible) {
+                this.clickEnemy();
+            }
         });
         this.name = data.name;
 
@@ -74,8 +78,15 @@ export class Enemy {
             console.log(this.name + " killed, getting " + this.killGold + " extra gold");
 
             // Update quest and stats
-            this.scene.enemyKilled();
+            this.scene.enemyKilled(this.name);
         }
     }
-
+    show() {
+        this.enemy.visible = true;
+        this.healthBar.show();
+    }
+    hide() {
+        this.enemy.visible = false;
+        this.healthBar.hide();
+    }
 }
--- a/source/enemy.js
+++ b/source/enemy.js
@@ -9,6 +9,7 @@ export class Enemy {
     scene;
     killGold;
     name = '';
+
     constructor(data){
         // Add enemy
         this.enemy = data.scene.add.image(data.x, data.y, data.name);
@@ -43,6 +44,7 @@ export class Enemy {
         this.killGold = data.killGold;
         this.scene = data.scene;
     }
+
     clickEnemy(){
         // Display hit
         let hitValue = Math.floor( Math.random()*2 ); // Currently just 50-50 chance 0/1
@@ -68,6 +70,7 @@ export class Enemy {
             _this.hitsplatText.visible = false;
         }, 200);
     }
+
     damageEnemy(damage){
         // Lower health and check status
         let isDead = this.healthBar.updateHealth(damage);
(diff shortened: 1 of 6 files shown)


source/level.js
```diff
@@ -201,44 +229,67 @@ export class Level extends Phaser.Scene{
...
         this.characterData.gold += addedGold;
         this.goldText.text = 'Gold: ' + this.characterData.gold;
     }
-    enemyKilled(){
+
+    enemyKilled(name){
         // Update kill quest score
-        if (this.characterData[this.background.name].enemiesKilled < this.killQuest) {
+        if (this.characterData[this.background.name].enemiesKilled[name] < this.killQuest) {
             this.characterData.totalEnemiesKilled++;
```
EthanVieira: Shouldn't this increment whether or not the quest is completed?

source/level.js
```diff
@@ -201,44 +229,67 @@ export class Level extends Phaser.Scene{
...
-            this.characterData[this.background.name].enemiesKilled++;
-            this.killQuestText.text = this.characterData[this.background.name].enemiesKilled + "/" + this.killQuest + " " + this.enemySettings.name + "s killed";
+            this.characterData[this.background.name].enemiesKilled[name]++;
+
+            let questCompleted = true;
```
EthanVieira, line 246: I would only set this to true in the if statement down below that checks for completion.

source/healthBar.js
```diff
@@ -58,4 +61,11 @@ export class HealthBar {
         return (wasDead);
     }
 
+    show() {
+        this.healthBar.visible = true;
+    }
+    
+    hide() {
+        this.healthBar.visible = false;
+    }
```
EthanVieira, line 70: Assuming the health bar won't need to be hidden separately, I would just trash these and inside the enemy show/hide function do this:
`show() {
        this.enemy.visible = true;
        this.healthBar.visible = true;
    }
`

source/enemy.js
```diff
@@ -9,13 +9,18 @@ export class Enemy {
...
         this.enemy = data.scene.add.image(data.x, data.y, data.name);
         this.enemy.setOrigin(.5,0).setDepth(2).setScale(.4);
         this.enemy.setInteractive();
+        this.enemy.visible = false;
         this.enemy.on("pointerup", ()=>{
-            this.clickEnemy();
+            // Check if it is the current enemy
+            if (this.enemy.visible) {
+                this.clickEnemy();
+            }
```
EthanVieira, line 23: For some reason using visibility for this rubs me the wrong way, but I can't think of a good reason why it wouldn't be fine, so it should be fine.

EthanVieira (changes requested):

source/enemy.js
CurtisGreen: I had a variable in there specifically for it but this ensures that it can never desync visuals and function

source/healthBar.js
CurtisGreen: Aight

source/level.js
CurtisGreen: This is part of that statement. I set this local variable to true then I loop through all of the enemies and switch it to false if any do not meet it. Then I check that Boolean value

source/level.js
CurtisGreen: Uhhhhhhh

source/level.js
EthanVieira: Yeah I was saying that more for readability. On second thought, just get rid of the variable and do the stuff that is in if (quest complete) inside the if before it (switch the < to >=)

source/healthBar.js
CurtisGreen: Oh, this would actually come out as `this.healthBar.healthBar.visible = false;`, still prefer it?

source/level.js
CurtisGreen: But it needs to come afterwards so you know that all of the enemies are defeated X # of times. For example 10 cows and 10 goblins defeated. Otherwise it would say completed when only one of them has been. I guess this is an intended functionality question

EthanVieira: We can leave it as is if you want - it is really more of a nit pick. I just don't like assuming it to be true until all the monster counts are proven to be >=.

source/healthBar.js
EthanVieira: Yikes. Leave it how you had it lol

EthanVieira (dismissed):

CurtisGreen:
I noticed a bug when I did those other changes. Fix was luckily 1 line change

EthanVieira (approved):

PR #12, opened 2019-10-25
Bgm

CurtisGreen:
1) BGM for load, tutorial island, and lumbridge
2) Settings button opens audio page
3) Volume can be adjusted and saves
4) Inventory button closes audio settings and opens inventory


EthanVieira:
How much work would it be to get an older sound settings image that doesn't have the bond image in the lower right corner?

CurtisGreen:
> How much work would it be to get an older sound settings image that doesn't have the bond image in the lower right corner?

What used to be there? If it was nothing I could edit it out

EthanVieira:
I can't find a picture of it, so I guess we will leave it as is.

EthanVieira (approved):

PR #14, opened 2019-11-09
Woodcutting

CurtisGreen:
Separated enemyLevel from level, added resource object and resourceLevel. Added progress bar instead of health bar for the tree


README.md
```diff
@@ -6,13 +6,13 @@
...
 
-Future ideas: See trello board
+[Google doc for future ideas](https://docs.google.com/document/d/14aIu6_-7gJhgkLuyItdyquUk115mISjRNwYPJmxmUm0/edit?usp=sharing)
 
 In progress:
                   
 * coming up with even bigger and brainier ideas - Ethan 
 * creating big brain item shop - Will
-* coming up with big brain ideas - Curtis
+* looking for his lost axe head - Curtis
```
EthanVieira, line 15: Absolutely unacceptable

source/enemyLevel.js
```diff
@@ -0,0 +1,133 @@
...
+
+    enemyKilled(name){
+        this.characterData.totalEnemiesKilled++;
+        // Update kill quest score
+        if (this.characterData[this.currentLevel].enemiesKilled[name] < this.killQuest) {
+            this.characterData[this.currentLevel].enemiesKilled[name]++;
+
+            let questCompleted = true;
+            this.killQuestText.text = "";
+            this.clickObjectMetaData.forEach((enemy, index) => {
```
EthanVieira, line 92: Opportunity to abstract out a method here since this is some identical code

EthanVieira (changes requested):
What is the purpose of the tree file that isn't a .webp?

source/enemyLevel.js
CurtisGreen: 1) They're the same, I mustve messed up copying it in there. 
2) Yeah I'm trying to think of a way I can abstract this but still keep the enemy related stuff in the enemyLevel class. Suggestions are welcome. I was mainly doing this because I thought that once we start actually getting resources from the trees and whatever stuff we do then we'd have to add more into this file. But I don't want to have a bunch of duplicate code

CurtisGreen: I didn't completely fix the duplicate code but I reduced it some by moving that one method back into Level.js and deleting ResourceLevel.js since it was tiny and functionally useless. I think later when resources are hashed out better it will be easier to clean this up fully.

EthanVieira (approved):

PR #15, opened 2020-01-26
Feature/world map

CurtisGreen:
**World map is now more intuitive:**

1) Centers around Tutorial Island and Lumbridge
2) Text moves with the map when you drag it
3) Text changes color based on what you have unlocked
4) Map can no longer be dragged off the screen
5) Map labels are now more readable and cover up the existing labels

**Other:**

1) Changed currentVolume default value to 3 from 0 as this will be used for the initial login music
2) Moved save data to a separate file so it can be edited without needing to modify level.js
3) Moved map constants to constants.js under a separate export


source/scenes/worldMap.js
```diff
@@ -16,23 +16,36 @@ export class WorldMap extends Phaser.Scene{
...
+        // Color links if they haven't been unlocked yet
+        let fillColor = 'yellow';
+        if (!this.characterData.TUTORIAL_ISLAND.questCompleted) {
+            fillColor = 'white';
+        }
+
+        // Tutorial Island
+        const tutorialIslandX = 1100;
+        const tutorialIslandY = 1070;
+        let tutorialIsland = this.add.text(tutorialIslandX, tutorialIslandY, 'Tutorial Island', {fill: 'yellow', fontSize: '20px', style: 'bold'}).setDepth(1);
```
zengk95: Should extract out the constants such as lumvridge and the text property object

source/scenes/worldMap.js
```diff
@@ -16,23 +16,36 @@ export class WorldMap extends Phaser.Scene{
...
-        // City text
-        let tutorialIsland = this.add.text(600, 360, 'Tutorial Island', {fill: 'white', fontSize: '20px'}).setDepth(1);
+        // Color links if they haven't been unlocked yet
+        let fillColor = 'yellow';
+        if (!this.characterData.TUTORIAL_ISLAND.questCompleted) {
+            fillColor = 'white';
+        }
+
+        // Tutorial Island
+        const tutorialIslandX = 1100;
```
zengk95: Why not just put this in the function
Or
Const [lumbx, lumby] =[10, 20]

zengk95:
Will this game ever be ready

source/scenes/worldMap.js
CurtisGreen: Moved into constants.js under a separate export

zengk95: I think you should move it back in less files the better

willadams05 (approved):

PR #17, opened 2020-02-08
Refactor

EthanVieira:
* Renamed and reorganized assets
* Standardized naming convention
* Removed some unnecessary stuff
* Reformatted code
   - Get Prettier and auto format before PRs or I will yell at you
* Updated Trello with TODOs found in code


CurtisGreen:
Improving Trello board and standardizing file names looks good. I'm not sure why it didn't realize it was a file rename instead of a file delete/add. If you know a way for it to recognize this I think it would make it much easier to review. If not oh well

src/scenes/level.js
```diff
@@ -0,0 +1,467 @@
...
+    // Show stats
+    let statColor = "white";
+    this.enemiesKilledText = this.add
+      .text(
+        20,
+        60,
+        "Enemies killed: " + this.characterData.totalEnemiesKilled,
+        { fill: statColor }
+      )
+      .setDepth(3);
```
CurtisGreen:
```
this.enemiesKilledText = this.add
      .text(
        20,
        60,
        "Enemies killed: " + this.characterData.totalEnemiesKilled,
        { fill: statColor }
      )
      .setDepth(3);
```

This is taking it a bit too far

EthanVieira: I'm letting Prettier do its thing. I believe if it can't fit all parameters on one line then it splits them up like this. I agree with what it is doing.

EthanVieira:
> Improving Trello board and standardizing file names looks good. I'm not sure why it didn't realize it was a file rename instead of a file delete/add. If you know a way for it to recognize this I think it would make it much easier to review. If not oh well

Agreed. I think the latest commit made github recognize this but not for all files. Not sure how to achieve this

CurtisGreen:
I think you'd have to start from where it was before and do it in separate commits, dw about it.

I noticed when I pulled that the source/ folder still existed though. Is that a git issue or did you forget to delete it

EthanVieira:
> I think you'd have to start from where it was before and do it in separate commits, dw about it.
> 
> I noticed when I pulled that the source/ folder still existed though. Is that a git issue or did you forget to delete it

Hmm well I just renamed it. I don't see the folder in my branch's git repo.

CurtisGreen:
Alrighty. I noticed the images for the shop do not load properly, I think it's a file path issue

EthanVieira:
> Alrighty. I noticed the images for the shop do not load properly, I think it's a file path issue

Fixed!

CurtisGreen:
> Fixed!

Looks good, I'm pretty sure the consumables part had that error before so we can add that to trello

CurtisGreen (approved):
Good stuff my dude

PR #18, opened 2020-02-14
Feature/map exit

CurtisGreen:
1) New button to exit the level to the main menu
2) New button to exit the minimap back to the level


src/scenes/level.js
```diff
@@ -221,6 +228,17 @@ export class LevelScene extends Phaser.Scene {
...
 
+        // Exit button
+        let exitButton = this.add
+            .image(this.width - 30, 0, "exit-button")
+            .setOrigin(0, 0)
+            .setDepth(2)
+            .setInteractive();
+        exitButton.on("pointerup", () => {
+            this.clearAutoClickers();
+            this.scene.start(CONSTANTS.SCENES.MAIN_MENU, this.characterData);
```
EthanVieira, line 240: This is great that you added this in. Can you update the audio to switch to the main menu theme? Or do you wanna do that in a different pr? Other than that it lgtm! https://www.youtube.com/watch?v=rR4n-0KYeKQ

CurtisGreen: It's really true about the big commits though, if they're small you get nitpicked.

I have it set the song before it changes the scene now as a temporary fix.
Gonna make a new task to:
1) Loop audio when song is finished
2) Under create, check what song is playing and play if it's different than the current. Should prevent the song from restarting when you go to the minimap then exit, also removes the need to set the song manually before changing to non-level scenes

EthanVieira (approved):

PR #19, opened 2020-02-16
Feature/stats panel

CurtisGreen:
**Added in this PR**
1) Added skills panel
2) Added combat stat level text and level formula
3) Damage now scales with combat level for the respective class

**Future tasks from this:**
1) Figure out how to make small text less fuzzy
2) Add stats for other skills like woodcutting
3) Discuss how/if we want to change how level is determined
4) Separate skills panel somehow, maybe parallel scene


src/enemy.js
```diff
@@ -1,4 +1,5 @@
 import { HealthBar } from "./ui/health-bar.js";
+import { CONSTANTS } from "./constants/constants.js";
```
EthanVieira: Unused?

src/scenes/level.js
```diff
@@ -506,4 +613,38 @@ export class LevelScene extends Phaser.Scene {
...
+        this.magicBottomText.text = level;
+        totalLevel += level;
+
+        this.totalLevelText.text = totalLevel;
+    }
+
+    calcLevel(xp, lv = 1) {
+        let currLvXp = 75*(Math.pow(1.104, lv));
+        if (xp > currLvXp) {
+            return (this.calcLevel(xp - currLvXp, lv+1));
```
EthanVieira: Should just invert the equation and ceil(?) the result (watch out for negative levels in xps in lvl 1 range). Also I'd prefer to use the actual xp equation even if it is more complicated.

src/scenes/level.js
```diff
@@ -252,6 +273,53 @@ export class LevelScene extends Phaser.Scene {
...
+            .image(560, 168, "skills-button")
+            .setOrigin(0, 0)
+            .setDepth(2)
+            .setInteractive();
+        this.skills.button.on("pointerup", () => {
+            this.showSkills(true);
+        });
+
+        // Skills text
+        this.attackText = this.add
```
EthanVieira, line 293: These are defaulted and shown whenever you move to a different area. They aren't updated until an enemy is clicked - it is never updated in woodcutting scene. These should display what is in character data and not defaulted to 1s.

EthanVieira (changes requested):

src/enemy.js
CurtisGreen: Oh yeah, I initially had the xp function under constants but apparently you can't reference yourself under an object like that so I couldn't do it recursively.

src/scenes/level.js
CurtisGreen: Inverting it would just check if you have the xp difference at the current level, not if you have the total xp necessary. This could be circumvented by storing the level and xp separately and resetting xp to 0 once you level up, but I'm pretty sure runescape shows total xp on hover

src/scenes/level.js
CurtisGreen: oops

Also I haven't done woodcutting yet, I was just prioritizing combat, wouldn't be too hard to add the text for this though

CurtisGreen: Added the text and xp

src/scenes/level.js
EthanVieira: Use the rewritten summation without the truncation error: https://runescape.fandom.com/wiki/Experience#Equations

CurtisGreen: Woops, will do when I'm home

CurtisGreen: For future self:
```
function calcLevel(xp, lv = 1) {
    let curLvXp = Math.floor(.25*(lv + 300*Math.pow(2, lv/7)));
    if (xp > curLvXp) {
        return (this.calcLevel(xp - curLvXp, lv+1));
    }
    else {
        return lv;
    }
}
```

CurtisGreen: Done

EthanVieira: If you use the L/8(L-1) + 75*(... ) + E(L) you should not need recursion.

EthanVieira: Oh you would actually need something more sophisticated. Alright I'm in favor of simple and stupid.

EthanVieira (approved):

EthanVieira (approved):

EthanVieira (approved):

EthanVieira:
Nice I can approve these ad infintum and the green checks will stack

EthanVieira (approved):

EthanVieira (approved):

EthanVieira (approved):

PR #21, opened 2020-02-22
Feature/varrock

CurtisGreen:
1) Added varrock
2) Added method to center on correct city when the world map is opened up
3) Enemy names are now capitalized

___
Fixed issue where cookie reset did not work properly because the data was done by a shallow copy rather than a deep copy


src/scenes/level.js
```diff
@@ -45,8 +45,8 @@ export class LevelScene extends Phaser.Scene {
     autoClickers = [];
     autoClickDps = 0;
 
-    // Character
-    characterData = saveData;
+    // Character (deep copy)
+    characterData = JSON.parse(JSON.stringify(saveData));
```
CurtisGreen: I initially wanted to use:
`characterData = Object.assign({}, saveData);`
But ironically it only works for the top level of data making it somewhat of a shallow copy lol.
aka: `characterData.characterClass` gets wiped but `characterData.TUTORIAL_ISLAND.enemiesKilled.rat` does not

EthanVieira (commented):
oops

src/save-data.js
```diff
@@ -1,4 +1,4 @@
-export var saveData = {
+export const saveData = {
```
EthanVieira: How does this work?

src/scenes/varrock.js
```diff
@@ -0,0 +1,28 @@
...
+            killQuest: 15,
+            background: {
+                name: "varrock",
+                path: "src/assets/backgrounds/VarrockBackground.png"
+            },
+            minimap: {
+                name: "varrock_map",
+                path: "src/assets/maps/VarrockMap.png"
+            },
+            clickObjects: [
```
EthanVieira, line 17: I'd like to see some other enemies here. City guards would be good. Maybe bears would be good.

src/save-data.js
EthanVieira: Ah I see now. It should probably be renamed to better indicate its purpose.

src/scenes/level.js
```diff
@@ -82,16 +82,15 @@ export class LevelScene extends Phaser.Scene {
...
-
         // Receive cookies if they exist
         if (characterData.hasCookies) {
             this.characterData = characterData;
         }
         // Otherwise, initialize character based on starting class
         else {
-            switch (this.characterData.characterClass) {
+            // Reset data
+            this.characterData = JSON.parse(JSON.stringify(saveData));
```
EthanVieira: Is this line needed? Either we initialize to null or remove this. Also I'd say the "Reset data" comment is a little misleading and the one on line 89 is sufficiently accurate

EthanVieira (changes requested):

src/save-data.js
CurtisGreen: The idea was to have the default/zeroed out data so rather than having to write `this.characterData.thing1 = 0; this.characterData.thing2 = 0` etc. you can just say curData = saveData and it resets it

src/scenes/varrock.js
CurtisGreen: I can do guards but aren't they too high of a level this early on?

src/scenes/level.js
CurtisGreen: It is needed unless you can think of a better way to do it. Resetting it to null will cause errors when you try to reference the data in the second level of the object. This needs to be here for the reset cookies button to work otherwise it will maintain the previous data

src/save-data.js
CurtisGreen: Do you think I should rename it to "defaultData"?

EthanVieira: Yea defaultData would be much better

src/scenes/varrock.js
EthanVieira: Yea probably but we don't have to be completely accurate with that kind of stuff

src/scenes/level.js
EthanVieira: I'm saying that aren't we already initializing to this on line 49?

EthanVieira: OH for the reset cookies button. I see i see. Can we initialize to null on line 49?

CurtisGreen: Yeah that'll work

src/save-data.js
CurtisGreen: Done

src/scenes/varrock.js
```diff
@@ -20,6 +20,12 @@ export class VarrockScene extends EnemyLevelScene {
                     path: "src/assets/sprites/DarkWizard.png",
                     maxHealth: 12,
                     killGold: 10
+                },
+                {
+                    name: "guard",
+                    path: "src/assets/sprites/VarrockGuard.png",
```
EthanVieira: I'm pretty sure this image is from future updates. Use the image you have for Falador guard for these. Much much much more nostalgic

EthanVieira (changes requested):

src/scenes/varrock.js
EthanVieira: Also please rename falador guard to just guard

CurtisGreen: That's what I was thinking but I opened up osrs and that's what they have rn. I can switch it though. Do you want to have the same guards for both Varrock and Falador?

EthanVieira: Imo it should be the same.

CurtisGreen: Aye aye cap'n

EthanVieira (approved):

PR #22, opened 2020-02-23
Separate panel scene

CurtisGreen:
1) Moved code related to the in-game UI (skills, sound, etc.) into a new file called dashboard.js. Creates a separate parallel scene so it can manage loading its own assets
2) Moved calcLevel under constants.js
3) Fixed issues with shop loading


src/resource.js
```diff
@@ -49,13 +49,13 @@ export class Resource {
...
-        let curLv = this.scene.calcLevel(curXp);
+        let curLv = calcLevel(curXp);
 
         // Increase xp
-        this.scene.updateSkillsText();
+        this.scene.dashboard.updateSkillsText();
 
         // Increase progress and check completion
-        let completed = this.progressBar.updateProgress(curLv);
+        let completed = this.progressBar.updateProgress(curLv);q
```
CurtisGreen: woops gimme a sec

CurtisGreen: Got it

src/constants/constants.js
```diff
@@ -71,3 +72,13 @@ export const MAP = {
         backgroundColor: "grey"
     }
 };
+
+export const calcLevel = function(xp, lv = 1) {
```
EthanVieira, line 81: Not sure if this is the appropriate place for this. Should we make a utils file for misc methods? Thoughts?

src/scenes/enemy-level.js
```diff
@@ -172,14 +177,28 @@ export class EnemyLevelScene extends LevelScene {
...
                 break;
             case CONSTANTS.CLASS.WARRIOR:
-                return this.calcLevel(this.characterData.skills.attack);
+                return calcLevel(this.characterData.skills.attack);
                 break;
         }
     }
+
+    // Show/hide text
+    showQuestText(show) {
```
EthanVieira: NIT: can we change the parameter name in both of these methods to isShown or something more indicative of a boolean?

src/ui/dashboard.js
```diff
@@ -0,0 +1,372 @@
...
+                audioButtonRow.push(audioButton);
+            }
+            // Save 2d array of buttons (3 x 5)
+            this.audio.audioButtons.push(audioButtonRow);
+        }
+        // Hide audio page on startup
+        this.showAudioSettings(false);
+    }
+
+    showSkills(show) {
```
EthanVieira: isShown for these as well

EthanVieira (changes requested):

src/constants/constants.js
CurtisGreen: Yeah that sounds good but I'm not sure what else would go there as of now

CurtisGreen: Gonna wait til we have more related utilities then I'll separate it out.

src/scenes/enemy-level.js
CurtisGreen: Got it

src/ui/dashboard.js
CurtisGreen: Done

EthanVieira (approved):

PR #23, opened 2020-02-29
Clean, full-window, world map

CurtisGreen:
**World map**
1) Replaced the world map with a label-less one
2) World map takes up the whole window instead of just the normal game dimensions

**Other**
1) Fixed Varrock mini-map not lining up
2) Changed enemy kill quest text to be added vertically 
    * Rather than being `enemy1 0/10 & enemy2 0/10` it's 
    * enemy1 0/10
    * enemy2 0/10
3) Updated phaser version and changed it to link to the minified version online rather than storing it locally
4) Removed unused typscript library


EthanVieira (approved):
Looks great!

PR #24, opened 2020-03-01
Feature/barbarian village

CurtisGreen:
**Barbarian Village**
1) Fulfilled everyone's need to hear some barbarianism

**Other**
1) Opening the world map in areas close to the edge of the map will no longer move the screen outside the boundaries
2) Used the "event" that I used for resizing the map to create a destructor for the level scene

image
image


EthanVieira (approved):
money

PR #26, opened 2020-03-23
Inventory

CurtisGreen:
**Inventory**
1) Inventory now exists, can left click or right click to use, examine, drop, etc.
2) Enemies have a 50% chance to drop bones (adjustable), trees drop logs
3) Item types (currently have Bones and Logs) can be added by adding their class in /ui/itemTypes.js and adding them as a drop in a level
4) Items are saved in cookies

**Other**
1) Changed prettier settings to increase width by a bit, felt a bit cramped
2) Moved autoclicker text to the middle left so it's not in the way of the inventory and made it gold
3) Moved item loading to the load.js scene rather than dynamically in the shop
4) Added subfolders like weapons under assets/items (that's why theres a bunch of changed files)

**Todo**
1) Make prayer and make burying bones actually increase prayer XP
2) Make items draggable to move around
3) Cleanup shop item class vs classes in itemTypes
4) Move more customization into the individual scenes and re-combine level and enemy-level, and maybe recombine resource/enemy classes to a generic "Target" class. 
5) Make right click menu close more consistently (happens when you move your mouse over the options text when exiting the box, this causes it to not register as a pointerout event)

*More todos*
6) Implement the rest of the weapon classes and add them into the item-manifest object


src/ui/itemTypes.js
```diff
@@ -0,0 +1,148 @@
...
+
+    switch (itemName) {
+        case "Bones":
+            return new Bones(itemData, matData, scene);
+            break;
+        case "Logs":
+            return new Logs(itemData, matData, scene);
+            break;
+    }
+}
```
CurtisGreen: I'm not sure how to do this with how the weapons are currently organized. In items.js, "Bones" and "Logs" are currently sitting directly under the parent ITEM whereas since there are multiple types of weapons, the data for other weapons is under ITEM.Weapon.Dagger for example. 

A couple ways for this to work, 
1) Case statement has all weapon names there and creates a Weapon class if it is listed
2) One of the input parameters to the function is the explicit path to the data, for example "Weapon.Scimitar" or "Weapon.Crossbow"

src/constants/items.js
```diff
@@ -1,23 +1,26 @@
...
-        DAGGER: {
+        Dagger: {
             name: "Dagger",
             cost: 50,
-            material: "SMITHINGMATS",
-            skill: "attack"
+            material: "ORE",
+            skill: "attack",
+            examineText: "Short but pointy"
         },
```
CurtisGreen, line 10: In case you were wondering, I changed these names so that they match up with the class, I plan to make a Weapon class later so might as well adjust it now.

I could have Weapon be the same but the subcategories like DAGGER be in all caps, but I think it might look off. Thoughts?

src/ui/inventory.js
```diff
@@ -0,0 +1,182 @@
...
+                .setInteractive()
+                .setDepth(5)
+                .on("pointerdown", () => {
+                    item[action]();
+                    this.menu.destroy();
+                });
+
+            options.push(actionText);
+            options.push(itemText);
+        });
```
CurtisGreen: You might notice in the right click menu that the capitalization is not consistent (use item vs Examine item). This is because when I dynamically create the text and action I use this syntax: `item[action]();` which is essentially item.use(); or item.bury();, etc. 

The simplest way to fix this would to change "use" to "Use" in both the item.actions[] array as well as the function name itself. Function names being capitalized is against convention however so I was hesitant to do this. lmk what you think

src/constants/items.js
```diff
@@ -28,23 +31,26 @@ export const ITEMS = {
...
         // Ranged Weapons
-        SHORTBOW: {
+        Shortbow: {
             name: "Shortbow",
             cost: 100,
-            material: "FLETCHINGMATS",
-            skill: "ranged"
+            material: "WOOD",
+            skill: "ranged",
+            examineText: "Short but affective"
```
EthanVieira: effective!

src/constants/items.js
```diff
@@ -72,20 +78,34 @@ export const ITEMS = {
...
         }
     },
-    CONSUMABLES: {}
+    Logs: {
+    	name: "Logs",
+    	cost: 5,
+    	material: "WOOD",
+    	examineText: "A number of wooden logs"
+    },
+    Bones: {
```
EthanVieira: Since I can't comment on individual files, I'll comment here. I'm pretty sure that image is of big bones and not bones. If monsters are gonna drop bones then we will want to discern between bone types.

src/constants/items.js
EthanVieira: Wouldn't we want them to be different than the class so we can easily tell the difference? Do we still want these constants separate from the class? I'm not sure if it is better to keep them here and retrieve them statically or to have instances of these attributes in each object.

src/scenes/lumbridge.js
```diff
@@ -19,13 +19,15 @@ export class LumbridgeScene extends EnemyLevelScene {
                     name: "cow",
                     path: "src/assets/sprites/Cow.png",
                     maxHealth: 8,
-                    killGold: 5
+                    killGold: 5,
+                    drops: [{item: "Bones", mat: "Normal", rate: .5}]
```
EthanVieira: Probably want to be consistent between mat and material

src/ui/inventory.js
```diff
@@ -0,0 +1,182 @@
...
+
+    constructor(scene, inventory) {
+        this.scene = scene;
+        this.playerItems = inventory;
+
+        // Update and show inventory on startup
+        this.playerItems.forEach(item => {
+            // Add blanks
+            if (!Object.keys(item).length) {
+                this.addToInventory({});
```
EthanVieira: Why exactly do we need to fill it up with blanks?

src/ui/inventory.js
```diff
@@ -0,0 +1,182 @@
...
+                    pointer.worldY > y + menuBox.height / 2 ||
+                    pointer.worldX < x - menuBox.width / 2 ||
+                    pointer.worldY < y - menuBox.height / 2
+                ) {
+                    this.menu.destroy();
+                }
+            });
+
+        // Add text options
+        // Have to use two separate texts per option for different clors
```
EthanVieira: colors!

src/ui/inventory.js
```diff
@@ -0,0 +1,182 @@
...
+    async checkForEmptySlot() {
+        for (let index = 0; index < this.playerItems.length; index++) {
+            if (!Object.keys(this.playerItems[index]).length) {
+                return index;
+            }
+        }
+        return -1;
+    }
+
+    createRightClickMenu(x, y, item, index) {
```
EthanVieira, line 74: We will probably want the background image for the right click be automatically sized based on the number of options

src/ui/inventory.js
EthanVieira: We will definitely want them to be capitalized. It would probably make more sense for a lot of the specific option stuff to be within an item class. I.e. the items would have a method that returns all of its options (the text representing and the respective functionality) and this will just generate the interactive image.

src/ui/inventory.js
```diff
@@ -0,0 +1,182 @@
...
+        }
+        else {
+            // Use this ugly method to see if obj is empty
+            if (this.curSelectedItemIndex >= 0 && Object.keys(this.inventory[this.curSelectedItemIndex]).length) {
+                this.inventory[this.curSelectedItemIndex].highlightItem();
+            }
+            this.curSelectedItemIndex = index;
+        }
+    }
+
```
EthanVieira, line 139: Eventually we'll want dragging/swapping functionality

src/ui/itemTypes.js
EthanVieira: I'm not sure we should generalize this. Each "entity" should be a class that will create/drop/instantiate other specific item objects directly. We should just call their constructors directly within those classes. I'm also more in favor of keeping classes to their own files.

EthanVieira (changes requested):

src/ui/itemTypes.js
CurtisGreen: So you think there should be a "Dagger" class rather than using a "Weapon" or "MeleeWeapon" class for it? I feel like that would generate a lot of duplicate code

src/ui/inventory.js
CurtisGreen: If i'm understanding correctly that's what this already does. 
The ItemTypes class defines the methods "drop" and "examine" because they exist for all items. Then each respective subclass defines an array of additional options such as "use", "bury", "wield", etc. Thinking about it now though, I think all items have a "use" options so I can probably move that.

Other than capitalization is there anything you would change for this?

src/constants/items.js
CurtisGreen: I'm pretty sure I took it off the wiki for bones (then ran it through an online converter because they love using webp instead of png), but I'll double check when I get home.

src/scenes/lumbridge.js
CurtisGreen: I thought I changed this to material.

src/ui/itemTypes.js
EthanVieira: The better the class hierarchy the less amount of duplicate code there will be. There should be classes for each of these: item -> weapon -> meleeweapon -> dagger -> bronze dagger. Allowing the leaf classes (bronze dagger here) to use any of its parents methods/attributes will minimize duplicate code. Drop/Examine can be defined much higher than other potential options. Examine can just pull from an examine text attribute that is required to be defined in the leaf.

src/constants/items.js
EthanVieira: Looks like the wiki page uses big bones in the larger image but normal bones in the smaller one for some reason.

src/ui/inventory.js
EthanVieira: See if you can get the below stuff within the single loop:
for optionText, optionMethod in item.options -> item.options returns hashmap of text : -> method
    generate one line of background image
    add text and set interactive with the method

Only have the cancel text/option defined here

src/ui/inventory.js
CurtisGreen: To maintain item locations on refresh. Currently I use the same function for populating on startup (where I know the correct index) and populating during run-time (where I don't know the correct index). 

I think an improvement I could do to resolve this would be to create two separate functions. One for inserting into inventory at a specific index and one that finds the first available index and calls the insert function on that index. Right now I have those two combined and so it requires me to make this additional operation. Can fix

src/ui/inventory.js
CurtisGreen: Yeah I'm trying to think of how to do this since I'm using a screenshot of the in-game right click menu. 
1) Take multiple screenshots of different sizes and load right-click-[#options]
2) Draw the menu ourselves using the Phaser.Graphics object

src/ui/inventory.js
CurtisGreen: Yeah that's the purpose of the methods in ItemType for changing their index and stuff. I've looked into how I'd do it and phaser has built in things called "drop zones" that allow you to drag and drop and it locks into a spot. 

When we do that we would call those methods plus update this variable.

src/ui/inventory.js
EthanVieira: Multiple screenshots seems like an ape-like thing to do but honestly whichever looks better works for me lol

src/ui/itemTypes.js
CurtisGreen: The methods that a bronze dagger would need are the same as a mithril dagger would, the only differences between them are data like damage and material name. These would be stored for all weapons so I don't see the benefit of adding a new class for each individual item.

src/ui/inventory.js
CurtisGreen: This could be done by moving the use, drop, and examine methods into ItemType.actions[]. Then in the child classes instead of creating the actions array you just push new actions to front.

src/constants/items.js
CurtisGreen: For the 2nd part we can see what we decide from our discussions. 

What do you mean by "Wouldn't we want them to be different than the class so we can easily tell the difference?", for this it would still be ITEMS.Weapon rather than just Weapon for the weapon class so I think that should be distinguishable enough. What would you suggest if we changed them?

src/ui/inventory.js
CurtisGreen: Also for the capitalization, I think I would do this instead of what I have now:
```
actions = [{text: "Use", func: "use"}, 
           {text: "Drop", func: "drop"}, 
           {text: "Examine", func: "examine"}];
...
let actionText = this.scene.add.text(x - 48, optionsY, action.text, FONTS.OPTIONS_MENU)
item[action.func]();
```

Adds just a little bit more into the ItemType class but allows for flexibility

src/ui/inventory.js
CurtisGreen: I can't think of any circumstance where you would need more than 4 so I think it might be the easiest and potentially the fastest way assuming the images are loaded on startup rather than dynamically. If you think there would be more than that i think would definitely look into how to draw it. For now I'll just add it as a task, not a priority rn imo

src/ui/itemTypes.js
EthanVieira: If for no other reason, do it for organization. It makes more sense to have actual items one will see in-game be the leaf classes than to have a "logical parent" be the leaf.

src/constants/items.js
EthanVieira: These are still constants so they should be capitalized. If we were pulling the attribute from a static weapon class then it should be Weapon.CONSTANT_ATTRIBUTE. By having it be ITEMS.Weapon then it makes the reader think we are pulling a class out of the ITEMS hat.

src/ui/inventory.js
EthanVieira: By ItemType class do you mean Item class? They should have a default implementation in the item class if we think all items will have those options (they should, although some items in RS have destroy instead of drop to prevent duplicates by re-picking up). That looks good for the capitalization.

src/ui/inventory.js
EthanVieira: If we have degradable items there would be a "check status" or w/e. Also in a bank or shop interface there could be a lot of options when right clicking.

src/ui/itemTypes.js
CurtisGreen: Alright, I have an idea for how this could be cleanly implemented but it would require all the leaf classes to be in one file:
Item-classes.js
```
export class BronzeDagger() {}
export class IronDagger() {}
...
```
Item-types.js
```
import * as ItemClasses from "./item-classes.js";

export function getItemClass(itemName, material, scene) {
    return new ItemClasses[material + itemName]();
}
```
I've tried similar things for classes and I'm pretty sure this will work and remove the need for a massive case statement. This doesn't allow for separate files for each class, but only allows for a separate file for the parent Item class vs all the children. What do you think?

Actually you might be able to do:
```
import BronzeDagger as ItemClasses.BronzeDagger...
import IronDagger as ItemClasses.IronDagger...
...
```

In which case you have the benefits of both, but a bigger header instead.

src/constants/items.js
CurtisGreen: Makes sense, we can remove this while moving it over into actual classes like we discussed though

src/ui/inventory.js
CurtisGreen: Yeah right now I have it called ItemType I think because there was already an Item class defined for use in the shop. I think i'll rename ItemType to Item when I make these changes but keep the other Item class for now. The shop stuff can be modified in a later PR.

src/ui/itemTypes.js
EthanVieira: Can you store the item classes within the drop tables? That would remove the need for getItemClass

src/ui/inventory.js
CurtisGreen: Yeah I think we should have a separate item class specifically for shop items because all shop items have the same buy options that aren't related to the items at all outside of their examine text.

src/ui/itemTypes.js
CurtisGreen: I think you can but there would still need to be some method to recreate items just based on text when you refresh from cookies. When you convert classes to JSON they lose their functions

EthanVieira: Could you store the constructor?

CurtisGreen: I'm not sure. Do you have any idea of what that would look like?

EthanVieira: Idk JS well enough, but if you cant directly story the constructor method then you might be able to use a static factory method to construct the objects

CurtisGreen: After looking into it, it's possible but the work we'd have to do for it and the amount of text data saved doesn't make it an attractive option.

I did find out that you can dynamically import which would reduce the need for all these lines of imports so it wouldn't be that bad.

bronze-dagger.js
```
export default class BronzeDagger() {}
```

item.js
```
export function getItemClass(item, material) {
    let item = await import("./items/" material +"-" + item + ".js"); // "./items/iron-dagger.js"
    return item.default();
}
```

CurtisGreen: I'll look into factory methods in js though

CurtisGreen: I decided to create a file that I called item-manifest that has the names and file paths for both the classes and the images. This is used in the getItemClass function to dynamically import the class. This also should speed up load times on startup once all the weapons have been implemented because it won't have to loop through as many objects to find the item/material pairings, it'll just run down this one list.

CurtisGreen:
Implemented all the changes we talked about. It will take a lot of manual data entry to create all the individual item classes for each weapon combination so I'm going to do those on a rolling basis as we add them into the game.

src/items/item-manifest.js
```diff
@@ -0,0 +1,12 @@
+export var itemManifest = {
+	NormalLogs: {
```
EthanVieira, line 2: I feel like the images associated with each item should be a part of its respective class

EthanVieira (changes requested):
Nice job overall!

src/items/item-manifest.js
CurtisGreen: It sorta is, the add.image() function uses the item's name as the image name. Also, you still need to make sure the name and path are consistent between load and use

CurtisGreen: Do you have an idea of what you would like changed?

Also, you should enable this to be loaded on github pages, it's in the settings and you should just be able to set master as the host branch and then ~30 minutes later it should be up and running. Then you can just go to EthanVieira.github.io/rs-clicker I think

EthanVieira: Eh it is good enough as is. Should be enabled as rsclicker.io now

EthanVieira (approved):

PR #28, opened 2020-04-02
Level restructuring

CurtisGreen:
Small PR, was planning on doing a restructuring of level.js, enemy-level.js, resource.js, and enemy.js to make future development more streamlined but it's hard to do when I'm not exactly sure what new development is going to look like.

What I did:
1) Moved stats related things into their own scene, display stats based on level type
2) Fixed volume not being maintained between level/main menu
3) Fixed shop using xp instead of level for item display
4) Fixed errors when you click before scene finishes loading
5) Added play link to readme


EthanVieira (changes requested):
I'm not seeing the stats get loaded in a new area until after a click and I don't think woodcutting xp and levels are working

CurtisGreen:
> I'm not seeing the stats get loaded in a new area until after a click and I don't think woodcutting xp and levels are working

I fixed the issue with woodcutting but i'm not seeing the issue with stats not loading until you click. Can you try to get a reproduction case for me?

EthanVieira:
> > I'm not seeing the stats get loaded in a new area until after a click and I don't think woodcutting xp and levels are working
> 
> I fixed the issue with woodcutting but i'm not seeing the issue with stats not loading until you click. Can you try to get a reproduction case for me?

Start a new game and stay at tutorial island until lvl 2 attack. Then go to lumbridge and before attacking an enemy look at the stats - I see lvl 1 attack until I click on the enemy.

CurtisGreen:
Added a listener to call the update function if the function is called before the scene is loaded.

EthanVieira:
Are autoclickers supposed to give xp when killing monsters? If so, I don't think they currently are. Also there are some bugs when buying an autoclicker -> clearing cookies -> weird things where autoclickers are still working.

CurtisGreen:
> Are autoclickers supposed to give xp when killing monsters? If so, I don't think they currently are. Also there are some bugs when buying an autoclicker -> clearing cookies -> weird things where autoclickers are still working.

No I wanted to encourage some clicking by making the attack skills only increase from it. What do you think?

Also I bought an auto clicker, exited to main menu, reset cookies, then played and the autoclickers were not saved. I'll try more cases but I'm not seeing the issue right now. If you can find a reproduction case for that too let me know

EthanVieira (approved):

PR #31, opened 2020-04-10
Target restructuring

CurtisGreen:
## Plan
~~Not ready to complete the PR for this but I wanted your feedback before I do more work on it.~~
Edit: Felt like going for it so I did

I'm not happy with how I have level, enemy-level, enemy, and resource. There's a lot of overlap between Enemy and Resource so I made a parent class called Target to be the generic object that you click on. I think Target is a better name than ClickObject that I'd sorta been using. 

### Idea
Next what I'd like to do is get rid of enemy-level, move some of the functionality into the target class and others are already in level.js but just need to be made generic. 

From there I would do what we are currently doing with items and make children classes for each type of enemy, tree, etc.

### Questions
Biggest questions I have are on how I implemented and named the target/enemy/resource classes. 
Currently Target requires the child to implement:
1) A progress bar (health bar, etc.)
2) getClickValue() 
3) onClick(hitValue)
4) onCompletion()

In the case of Resource there is nothing that needs to be done on completion outside of getting the resource itself which is already done in the Target class.

### Target Restructuring
1) Implemented Target class
    * GiantRat -> Enemy -> Target
     * NormalTree -> Resource -> Target
2) Extended AutoClicker class: 
     * HiredBowman -> AutoClicker
3) Removed EnemyLevel and moved functionality into Enemy class and Level where appropriate

### Other
1) Temporarily autoclickers will now work on both trees and enemies due to them being made generic, this will be changed later when we make autoclicker additions
2) Fixed shop images for the webserver, once again local version was ignoring capitalization mismatches whereas the hosted version does not


src/scenes/level.js
```diff
@@ -90,9 +102,12 @@ export class LevelScene extends Phaser.Scene {
...
 
-        // Click object (target)
-        this.clickObjectMetaData.forEach(target => {
-            this.load.image(target.name, target.path);
+        // Targets
+        this.targetMetaData.forEach(target => {
+            let test = new target(this);
+            test.images.forEach(image => {
+                this.load.image(image.name, image.path);
+            });
```
willadams05: What is the extra test variable doing? Are the objects stored in targetMetaData not already target objects?

CurtisGreen: Rip forgot to change that. Interestingly enough they are uninitialized objects, until they are called with the "new" keyword the class is stored as what seems like text and its member attributes cannot be accessed until you initialize it.

I would like to add the objects into `targets[]`  here but in their constructor they call the `this.add.image()` function. I don't think you can do this during in the `preload` function so I just load the images here and then later on create the actual targets. 

I can think of 2 ways to address this:
1) just renamed test to temp and not worry about it
2) Move the `this.add.image()` into a separate create function under the Target class. Add the targets into the array here and then again in the create function of the Level class call each target's create function

src/scenes/level.js
```diff
@@ -255,6 +276,44 @@ export class LevelScene extends Phaser.Scene {
...
+                ) {
+                    questCompleted = false;
+                }
+                // Set as complete if all passed on last index
+                else if (questCompleted && index == this.targetMetaData.length - 1) {
+                    this.characterData[this.currentLevel].questCompleted = true;
+                    console.log("Quest complete!");
+                }
+            });
+        }
```
willadams05, line 277: What's the logic behind this function?

If you're just trying to check whether or not the quest has been completed, wouldn't something like this work: 

enemieskilled[name]++
if(enemiesKilled[name] == this.killQuest)
    characterData[currentLevel].questCompleted = true
    console.log("Quest complete")

What is the purpose of looping through each enemy in the targetMetaData object? Am just confused

src/scenes/barbarian-village.js
```diff
@@ -15,22 +16,7 @@ export class BarbarianVillageScene extends EnemyLevelScene {
...
-                },
-                {
-                    name: "female_barbarian",
-                    path: "src/assets/sprites/FemaleBarbarian.png",
-                    maxHealth: 20,
-                    killGold: 20,
-                    drops: [{item: NormalBones, rate: .5}]
-                }
-            ],
+            targets: [Barbarian],
```
willadams05, line 19: This is nice

src/scenes/lumbridge-trees.js
```diff
@@ -14,20 +14,7 @@ export class LumbridgeTreeScene extends LevelScene {
...
-                    drops: [{item: NormalLogs, rate: .5}]
-                },
-                {
-                    name: "tree2",
-                    path: "src/assets/sprites/Tree2.webp",
-                    neededClicks: 5,
-                    drops: [{item: NormalLogs, rate: .5}]
-                }
-            ],
+            targets: [NormalTree],
```
willadams05, line 17: Again, this is nice

src/targets/resource.js
```diff
@@ -0,0 +1,33 @@
+import { ProgressBar } from "../ui/progress-bar.js";
+import { CONSTANTS, calcLevel } from "../constants/constants.js";
+import { Target } from "./target.js";
+
+export class Resource extends Target {
+    skill;
```
willadams05, line 7: What is this lol

willadams05: Oh this is how you declare globals in js? Gross

willadams05 (dismissed):
I like the refactor, having the Target and Enemy classes is a nice abstraction

src/scenes/level.js
willadams05: I like the idea of the targets[] array, but if this is the only place you'd use it then it's fine to leave it as-is

src/scenes/level.js
CurtisGreen: `killQuest` is an int for the number of enemies needed to be killed to finish the level. If there are multiple enemies you need to kill `killQuest` # of them for each. 

This loops through and checks if you have killed the needed amount for all enemies, if so you have passed the level.

src/targets/resource.js
CurtisGreen: Do you mean export or what? Export just means other files can import that var/function/class. So sorta?

src/scenes/level.js
CurtisGreen: I tried adding it here then creating it later but it threw some other errors because of some other dependencies so I'm just gonna rename it and leave it as is I think.

src/targets/resource.js
willadams05: No just saying "skill;". I'm assuming that's how you declare a global without initializing in js

src/scenes/level.js
willadams05: So you can only beat a level once you've killed a certain number of each mob? Or a total number of monsters killed?

src/targets/resource.js
CurtisGreen: That's one way to declare a class member variable, it's not global. You can technically declare and use by just doing `this.varName = 0` and then using it somewhere, but I do it this way to show more clearly what vars we're using

src/scenes/level.js
CurtisGreen: If the killQuest is 10 and the enemies are a cow and a goblin, you need to kill 10 cows and 10 goblins

src/targets/resource.js
CurtisGreen: Or you could do `skill = {};` or `skill = "";` or whatever type it is

willadams05 (approved):

PR #32, opened 2020-04-17
Update constants.js

luckyduck728:


CurtisGreen:
Classes are being removed, see https://github.com/EthanVieira/rs-clicker/pull/33

PR #33, opened 2020-04-19
Equipment

CurtisGreen:
## Equipment

1. Added Equipment class that inherits from Item
1. Added equipment panel
2. Player now starts with Bronze Dagger, Shortbow, and Staff
2. XP gained and damage dealt now depends on weapon equipped
3. Changed damage and accuracy formula so that it matches what runescape uses
4. Added stats to all current enemies
5. Changed to use runescape icons rather than the 3d model pictures, moved existing images under the assets/items/models folder
6. Widened the right-click menu so "Bronze Dagger" doesn't go off the menu. Will need to make it dynamic later.

## Other
1. Fixed issue where resetting cookies did not set the current level
2. Fixed issue where wooducutting level would not be updated


willadams05 (approved):
Seems like cool stuff. Where did you get the stats for everything from? RSWiki?

CurtisGreen:
> Seems like cool stuff. Where did you get the stats for everything from? RSWiki?

Yeah they have the stats for all equipment/enemies in nice tables like here:
https://oldschool.runescape.wiki/w/Bronze_sword

PR #37, opened 2020-05-07
Stackable items

CurtisGreen:
## Other
1. Fixed issues that happened with online merge of previous two commits (one added images one deleted them)
1. Fixed pesky bug where equipping an item after purchasing would duplicate it (multiple instances of the inventory object containing the items were hanging around)
1. Added .prettierignore file to ignore fonts.css file since it messes it up
1. Fixed mute button, copying an array in js will do a shallow copy, copying values individually does a deep copy


src/items/item.js
```diff
@@ -57,6 +60,19 @@ export class Item extends ClickableObject {
                 }
             });
         this.displayHeight = this.sprite.displayHeight;
+
+        // Add text in top left for stackable items
+        this.numItemsText = this.scene.add
+            .text(x - 15, y - 15, this.numItems, {
+                font: "10px runescape",
+                fill: "orange",
```
willadams05, line 68: Mentioned on league. 1-99,999 = yellow, 100k-9999k = white, 10m+ = green

fonts.css
```diff
@@ -1,17 +1,17 @@
...
-    url("./src/assets/fonts/runescape.svg") format("svg"), /* Legacy iOS */
-    url("./src/assets/fonts/runescape.ttf") format("truetype"), /* Safari, Android, iOS */
-    url("./src/assets/fonts/runescape.woff") format("woff"), /* Modern Browsers */
-    url("./src/assets/fonts/runescape.woff2") format("woff2"); /* Modern Browsers */
-  font-weight: normal;
-  font-style: normal;
+    font-family: "runescape";
+    src: url("./src/assets/fonts/runescape.eot"); /* IE9 Compat Modes */
+    src: url("./src/assets/fonts/runescape.eot?#iefix") format("embedded-opentype"),
+        /* IE6-IE8 */ url("./src/assets/fonts/runescape.otf") format("opentype"),
```
EthanVieira: Revert this change.

src/items/item.js
```diff
@@ -112,7 +128,12 @@ export class Item extends ClickableObject {
 
     drop() {
         console.log("Drop", this.name);
-        this.destroy();
```
EthanVieira: When you drop a stackable item, it should drop the whole stack - like it does in runescape. Imagine trying to drop thousands of one item.

src/ui/dashboard.js
```diff
@@ -118,6 +118,10 @@ export class DashboardScene extends Phaser.Scene {
                 this.inventory.obj.showInventory(true);
                 this.inventory.button.setAlpha(0.1);
             });
+        // Clear out and reinstatiate inventory
```
EthanVieira: *reinstantiate?

src/ui/dashboard.js
```diff
@@ -318,6 +322,11 @@ export class DashboardScene extends Phaser.Scene {
 
                 this.equipment.obj.showEquipment(true);
             });
+
+        // Clear out and reinstatiate equipment
```
EthanVieira: Same spelling here

src/ui/inventory.js
```diff
@@ -61,12 +63,32 @@ export class Inventory {
...
         for (let index = 0; index < this.playerItems.length; index++) {
-            // Check if object exists
-            if (!Object.keys(this.playerItems[index]).length) {
+            let itemExists = Object.keys(this.playerItems[index]).length;
+
+            // Check if it can stack with other items
+            if (
+                itemExists &&
+                item.stackable &&
+                this.playerItems[index].item == item.item &&
```
EthanVieira, line 72: Can we simplify this by checking for the name?

src/ui/inventory.js
```diff
@@ -61,12 +63,32 @@ export class Inventory {
...
+                // Update it in the cookies
+                this.playerItems[index].count += item.numItems;
+
+                return true;
+            }
+            // Check if slot is empty
+            else if (!itemExists) {
                 this.addToInventoryAtIndex(item, index, createSprite);
                 return true;
             }
```
EthanVieira, line 95: Not sure how JS works but should we explicitly return false here?

EthanVieira (changes requested):

fonts.css
CurtisGreen: Prettier did this automatically

src/ui/dashboard.js
CurtisGreen: woops

src/ui/inventory.js
CurtisGreen: No because it's checking through the cookies which has the item/type and doesn't have the "name" only so that's why I check both. You also need to check if it exists or you'll get errors on empty inventory slots

src/ui/inventory.js
CurtisGreen: I'm returning false specifically when the item cannot be added because the inventory is full. Which I do at line 76/98. Not sure what you're talking about here

fonts.css
CurtisGreen: Do you want to revert it and then make sure prettier doesn't edit this file in the future or what?

src/items/item.js
CurtisGreen: I'm considering replacing the "drop" command with "sell". What do you think about that? Also I can make it so when the item count is > 1 it can be sell 1 or sell all. Or do you think this would overinflate the user's gold (which is already pretty inflated)

src/items/item.js
CurtisGreen: Do you want yellow or the orange that I'm using rn?

fonts.css
EthanVieira: I want the comments to be where they should

src/ui/inventory.js
EthanVieira: I was hoodwinked by the hidden code - still not used to github prs

src/items/item.js
EthanVieira: Only giving gold on monster death would help solve the inflation. Sell should probably only be when the shop is open

fonts.css
CurtisGreen: Kk i'm going to revert this and add *.css to the newly added .prettierignore file

src/items/item.js
CurtisGreen: Not sure what happened to the reponse I did. But right now the shop doesn't show inventory when it's open, if we want to sell only when the shop is open we would need to redesign the shop. If you have a good idea for a layout I can implement it. Only reason I thought of replacing "drop" with "sell" is because I don't see any reason why the user would want to drop something, unless we never want the user to sell anything

src/items/item.js
CurtisGreen: I just tried the yellow and I think orange looks better unless you have a specific hex value of yellow you want. I implemented the white + k/green + m with 1 decimal place after each. Ex: 1.2k, 5.1m. Lmk if that's what you were thinking of

src/items/item.js
EthanVieira: aight we can change to sell, but if we do that then might want a sell X item for stackable items

EthanVieira (approved):

PR #38, opened 2020-05-14
More store options

CurtisGreen:
## Clan
image
image

1. Rearranged the shop, making the window bigger and adding the option for clan members
    1. Clicking on the text in the shop adds them to the clan
1. Added the clan chat panel that shows all clan members
1. Scrollbar now takes height as a parameter
    1. Also made it work much more consistently in general

## Future
1. This part is just visual, need to make clan member classes that function as autoclickers
1. Make leveling up clan members with the relevant icon next to their name
1. Currently there is a bug where if you switch between shop tabs before it has finished loading it will show some of the items in the other tab. This generally only happens on the first time you open the shop since nothing has been cached yet


src/ui/scroll-window.js
```diff
@@ -12,93 +32,153 @@ export class ScrollWindow extends Phaser.Scene {
...
-                .image(width, scrollHeaderHeight, "scroll-button")
-                .setDepth(2)
-                .setOrigin(0, 0);
+    addObjects(data) {
+        if (data.objects.length > 0) {
+            this.refresh(data);
+        }
+    }
+
+    refresh(data = this.data) {
```
EthanVieira, line 47: Do you have plans to be able to click and drag the scroll button?

CurtisGreen: Yes, I made a card for it but didn’t think it was very high priority. I’ve already done drag for the world map though so maybe if I refer back to that I can get it done today in this PR

CurtisGreen: Alright, you can now drag it and use the top/bottom buttons to move it as well.

EthanVieira (approved):

PR #39, opened 2020-05-16
Attack animation

CurtisGreen:
## Attack animation
1. Added animations that vary for attack type
    1. Unarm is fist
    1.  Magic is fire bolt
    1. Ranged is bronze arrow
    1. Melee weapon depends on crush/slash/stab weapon type
1. Removed character select screen and character on the bottom left
    1. Had to edit the backgrounds of tutorial island/lumbridge because they were blocking some cows/rats that look tiny in comparison lol
1. Can't really screenshot so you'll have to run it

## Other
1. Added "Click DPS" stat, averages damage over the past second
1. Added the ability to change the offset for target/health bar only
    1. Offset Giant Rat, Cow, and Goblin to look better
1. Changed pretty much all fonts to be runescape style
1. Took a little JPEG out of the rsc logo
1. Updated readme pictures so you can see example screenshots on this branch


src/scenes/level.js
```diff
@@ -275,4 +237,112 @@ export class LevelScene extends Phaser.Scene {
...
+            scale = 1,
+            curve = 0,
+            alpha = 1;
+
+        // Get current weapon image
+        let curWeapon = this.dashboard.equipment.obj.equipment.WEAPON;
+        if (Object.keys(curWeapon).length) {
+            // Set animation based on current weapon
+            switch (curWeapon.skill) {
+                // Melee weapons use the weapon model image
```
EthanVieira: I'm all for explanations in the comments but try to be wary of commenting something that should be self-evident from reading the code

src/targets/target.js
```diff
@@ -53,6 +61,9 @@ export class Target extends ClickableObject {
     }
 
     clickTarget() {
+        // Click animation
```
EthanVieira: Right now we attack the trees with whatever is equipped. We should have a unique animation for that I'm guessing. Also might be good to only allow woodcutting after an axe has been purchased -> or just start with bronze as you do in rs tutorial.

Also looks like the axe animation hits the object with the wrong side of the axe. Honestly I don't care if that is how it is, but it does look a little weird and might not be how you intended

src/ui/stats.js
```diff
@@ -34,55 +38,53 @@ export class StatsScene extends Phaser.Scene {
...
-                    fill: statColor,
-                }
+                FONTS.STATS
             )
             .setDepth(3);
         this.autoClickDpsText = this.add
-            .text(20, 120, "AutoClicker DPS: " + this.autoClickDps, {
-                fill: statColor,
-            })
+            .text(20, 120, "AutoClicker DPS: " + this.autoClickDps, FONTS.STATS)
```
EthanVieira, line 71: idk what is causing it but there is overlap in the text on the Lumbridge woodcutting level

EthanVieira (changes requested):
There's something hilarious about this and I love it. Nice job

src/targets/target.js
CurtisGreen: Yeah I plan on doing a couple things relevant to that. 
Like make it so you can’t chop a tree w/o an axe. Do you think they should start with an axe or have to purchase one?
I also haven’t implemented level requirements yet so I’ll need to work on that, probably at the same time.

For the axe swing I tested a few items but not that one so maybe I’ll just mirror the image

EthanVieira: I'd say let them start with one

CurtisGreen: Okay you start with an axe now and you can only chop down trees if an axe is equipped. Also I flipped the image if it's an axe so it looks right

src/ui/stats.js
CurtisGreen: Forgot to add the visibility setter in my show/hide function for the new DPS tracker. Fixed

src/scenes/level.js
CurtisGreen: Removed that one comment

EthanVieira: Happens a good bit elsewhere but I don't care enough to force change. Just something to be cognizant of

CurtisGreen: Alright

EthanVieira (approved):

PR #40, opened 2020-05-19
absolute-path-fixes

CurtisGreen:
## Absolute path fixes
Currently there are two errors on the live server that are not reproducible locally, the lesser demon on the loading screen and dynamically importing the class of an item from the shop. The first error doesn't affect anything but the item import one breaks the shop.

After looking into it I think it's because these two paths are absolute instead of being relative so I've changed them to be such. The local version would be `localhost:8000/src/...` but the github pages one is `ethanvieira.github.io/rs-clicker/...` so I think it tries to access `ethanvieira.github.io/src/` instead of `ethanvieira.github.io/rs-clicker/src/`

I also ran prettier because I forgot to on the last commit of the last PR.

### DIFF
--- a/src/scenes/load.js
+++ b/src/scenes/load.js
@@ -15,7 +15,7 @@ export class LoadScene extends Phaser.Scene {
                     {
                         type: "image",
                         key: "lesser-demon",
-                        url: "/src/assets/sprites/LesserDemon.png",
+                        url: "src/assets/sprites/LesserDemon.png",
                     },
                 ],
             },
--- a/src/scenes/shop.js
+++ b/src/scenes/shop.js
@@ -239,7 +239,7 @@ export class ShopScene extends Phaser.Scene {
                 if (itemManifest[item].type == itemType) {
                     // Get item class
                     let path = itemManifest[item].classPath;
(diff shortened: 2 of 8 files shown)


src/scenes/load.js
```diff
@@ -15,7 +15,7 @@ export class LoadScene extends Phaser.Scene {
                     {
                         type: "image",
                         key: "lesser-demon",
-                        url: "/src/assets/sprites/LesserDemon.png",
+                        url: "src/assets/sprites/LesserDemon.png",
```
CurtisGreen, line 18: Here's 1

src/scenes/shop.js
```diff
@@ -239,7 +239,7 @@ export class ShopScene extends Phaser.Scene {
                 if (itemManifest[item].type == itemType) {
                     // Get item class
                     let path = itemManifest[item].classPath;
-                    let itemClass = await import("/src/items/" + path);
+                    let itemClass = await import("../../src/items/" + path);
```
CurtisGreen, line 242: Here's the other

EthanVieira (approved):

PR #41, opened 2020-05-23
Autoclicker rework

CurtisGreen:
## Autoclicker Rework
Autoclickers changed to use the more in-theme clan member functionality. Bunch of other small QOL changes and bugfixes

1. Buy autoclicker 50g text has been removed
1. Clan members bought in the store now function as actual autoclickers
1. Clan member text can be hovered over to view the cost and description

### Fixes
1. I think I fixed an issue where audio would sometimes not play on first launch

### Chat window
1. Enemies now have their max health shown on examine
1. Item descriptions in the shop only show while hovering, no longer persist when cursor moves away
1. Added placeholder player name to examine chat window

### Buttons
1. Added button class that can be used for callbacks without needing to setAlpha(.1) or background text shenanigans (it just sets an x/y, width/height area as interactive)
1. Replaced shop background text button with this type of button
1. Made the logout button interactive
1. The "delete cookies yes/no" buttons on the settings page now use the button area rather than the text as the callback (so it's easier to click)

### Other other
1. Item class importing has been moved to the loading screen so it only happens one time not every time you change scenes
1. Added "Attack" and "Chop" to the top of the options on the right click menu for enemies and trees respectively
1. Equipping an item now grays out the background behind it


src/auto-clickers/auto-clicker.js
```diff
@@ -1,42 +1,128 @@
...
+        this.cost = data.cost;
         this.dps = data.dps;
         this.level = data.level;
-        this.name = name.type;
+        this.name = data.name;
 
-        // Damage every .1 second
-        this.damageInterval = 0.1;
-        this.timer = setInterval(this.clickTarget.bind(this), this.damageInterval * 1000);
+        // Damage every .5 seconds
```
EthanVieira: Isn't it still .1 seconds?

src/auto-clickers/auto-clicker.js
```diff
@@ -1,42 +1,128 @@
...
+    async buy() {
+        this.scrollWindow.characterData.gold -= this.cost;
+        if (this.dashboard == undefined) {
+            this.dashboard = this.scrollWindow.scene.get(CONSTANTS.SCENES.DASHBOARD);
+        }
+        let newMember = await getAutoclickerClass(this.name, this.dashboard);
+        this.dashboard.clan.obj.addClanMember(newMember);
+    }
+
+    examine(isShop) {
```
EthanVieira, line 95: This goes for all examines - not just this one - but the right click menu should close when examine is clicked.

src/items/equipment.js
```diff
@@ -45,18 +47,22 @@ export default class Equipment extends Item {
...
+                }
 
-            this.actions[0] = { text: "Unequip", func: "unequip" };
-            this.equipped = true;
-            this.scene.equipment.obj.equipItem(this);
+                this.actions[0] = { text: "Unequip", func: "unequip" };
+                this.equipped = true;
+                this.scene.equipment.obj.equipItem(this);
+            } else {
+                console.log("Not high enough level to equip that.");
```
EthanVieira, line 64: Might want to put something on the trello board to make a logging level that prints to the chat window. This is something that would go there

src/items/equipment.js
```diff
@@ -85,4 +89,15 @@ export default class Equipment extends Item {
         super.highlightItem();
         console.log("use", this.name);
     }
+
+    // TODO: be able to have multiple different required levels for different skills
+    checkRequiredLevel() {
+        let skill = this.skill.toLowerCase();
+        let level = calcLevel(this.scene.characterData.skills[skill]);
+        if (level >= this.requiredLevel) {
```
EthanVieira: return (level >= this.requiredLevel);
😄

src/items/get-item-class.js
```diff
@@ -0,0 +1,8 @@
+import { itemManifest } from "./item-manifest.js";
```
EthanVieira: Do we need a whole file for this?

src/scenes/audio.js
```diff
@@ -73,10 +76,14 @@ export class AudioScene extends Phaser.Scene {
...
+        if (this.sfx != undefined && this.sfx.isPlaying) {
+            this.sfx.stop();
+            this.currentSong.pause();
+        }
+        this.sfx = this.sound.add(audioName);
+        this.sfx.setVolume(this.currentVolume[SFX] / 4);
+        this.sfx.play();
+        this.sfx.once("complete", () => {
             this.currentSong.resume();
         });
```
EthanVieira, line 89: Idk if these are the changes that caused this or if it was preexisting, but if you delete cookies, all subsequent volume settings will not be saved and it will always reset to 2/4 when switching scenes.

src/scenes/shop.js
```diff
@@ -182,7 +183,7 @@ export class ShopScene extends Phaser.Scene {
             goldText = gold / 1000 + "k";
         } else if (gold > 10000000) {
             goldText = gold / 1000000 + "M";
-            fill = "#06c663";
+            color = "#06c663";
```
EthanVieira, line 185: Unrelated, but could we move the cash stack location? Kinda looks like the Bot price. Maybe the bottom right corner?

src/targets/resources/normal-tree.js
```diff
@@ -1,18 +1,17 @@
-import { Resource } from "../resource.js";
+import { Tree } from "./tree.js";
 import NormalLogs from "../../items/logs/normal-logs.js";
 
-export class NormalTree extends Resource {
+export class NormalTree extends Tree {
```
EthanVieira, line 4: Idk if this caused it, but can't seem to chop trees. Autoclicker still works on them though

src/ui/chat-window.js
```diff
@@ -298,6 +327,17 @@ export class ChatScene extends Phaser.Scene {
                         this.enemyStatText[varName].text = object[varName];
                         this.enemyStatText[varName].visible = true;
                     }
+
```
EthanVieira, line 330: Enemy stat defenses aren't formatted nicely. Examine a cow and you'll see what I mean

src/ui/clan.js
```diff
@@ -97,9 +99,9 @@ export class Clan {
             this.scene.currentPanel = CONSTANTS.PANEL.CLAN;
         }
         this.scrollWindow.setVisible(isVisible);
-        this.clanMembers.forEach((member) => {
-            member.visible = isVisible;
-        });
+        // this.clanMembers.forEach((member) => {
```
EthanVieira: commented stuff?

src/ui/equipment.js
```diff
@@ -11,11 +11,19 @@ export class Equipment {
     equipment = {
         WEAPON: {},
     };
+    bg = {
```
EthanVieira: Can we get a better name for this?

EthanVieira (changes requested):
Dno how it will be in the future but might want to combine the same type of autoclicker in the chat at some point. Or maybe give them distinct names.

src/auto-clickers/auto-clicker.js
CurtisGreen: Lol yeah. I think I liked .1s better so i reverted it but didn't update the comment

src/auto-clickers/auto-clicker.js
CurtisGreen: How would you close the examine window then? Currently I have it so closing the right click menu closes the examine window as well

src/items/get-item-class.js
CurtisGreen: Can be put into utilities maybe

src/scenes/shop.js
CurtisGreen: I'll test it out later and see if I can find somewhere it looks good

src/targets/resources/normal-tree.js
CurtisGreen: You need an axe equipped. Does it still not work w/o an axe?

src/ui/chat-window.js
CurtisGreen: Yeah it's double digit and negative so it goes over. There's not a lot more space to work with there so suggestions welcome

src/ui/clan.js
CurtisGreen: Hmmmm

src/ui/equipment.js
CurtisGreen: bg = background. Do you mean you think it should be background instead or totally different?

src/scenes/shop.js
CurtisGreen: Added a trello story

src/targets/resources/normal-tree.js
EthanVieira: Ah ok in RS it didn't matter whether or not it was equipped. If we keep it that way we will prob want a log telling the player they need it equipped

src/ui/chat-window.js
EthanVieira: Use a second line!

src/ui/equipment.js
EthanVieira: bgWhenEquipped or something like that

src/auto-clickers/auto-clicker.js
EthanVieira: If we change examines to be a log to chat then it would be fine

CurtisGreen: Yeah, do you have any good ideas on where the close/open chat button would be? I think the chat log would open automatically when something is appended to it. But you should be able to open-close it at will other times

src/targets/resources/normal-tree.js
CurtisGreen: Yeah I have it in trello as a task to check for that. I'll probably do it at the same time as I add a check for the axe level

src/ui/chat-window.js
CurtisGreen: You mean have the numbers under the text?
like:
```
Attack  Strength  Defense  Magic  Ranged
1       1         1        1         1
Melee                      Magic  Ranged
-15                          0         0
...
```
Would basically double the line count, I don't think that'd fit:
image

src/ui/equipment.js
CurtisGreen: equipmentSlotBg?

CurtisGreen:
> Dno how it will be in the future but might want to combine the same type of autoclicker in the chat at some point. Or maybe give them distinct names.

Yeah I plan to use the ranking system so that ones you already have bought are upgraded instead. I'll use the clan icon next to them to denote upgrades

src/scenes/audio.js
CurtisGreen: Looks like it's pre-existing. I'll check it out

CurtisGreen: I think it's a visual bug rather than an audio one

CurtisGreen: It's both. I just fixed the visual issue of it, but the source still remains. It seems that when the cookies are reset the characterData pointer owned by everyone gets fractured. I can just put this.characterData = characterData at the top of every create function but this wouldn't work for the audio scene since it persists. I'll need to think of a way to propagate the change to everyone when cookies are reset. Either that or I'll need to make the characterData globally accessible through some getCharacterData() function.

CurtisGreen: Ah, I think since the audio scene persists it doesn’t get the updated object and it keeps the old pointer

CurtisGreen: Ah, I think since the audio scene persists it doesn’t get the updated object and it keeps the old pointer

src/ui/chat-window.js
CurtisGreen: Fixed it here
https://github.com/CurtisGreen/rs-clicker/pull/6

src/auto-clickers/auto-clicker.js
CurtisGreen: Addressed here:
https://github.com/CurtisGreen/rs-clicker/pull/6

PR #42, opened 2020-06-02
Fletching

CurtisGreen:
Note: I merged in the autoclicker changes to sort out some conflicts so review that one first

## Fletching
Knife + 50 logs = Shortbow
image

* Equipment now stacks so when you're fletching it doesn't fill up your inventory

## Skill Info
image

image


src/auto-clickers/auto-clicker.js
```diff
@@ -1,42 +1,128 @@
...
+        this.cost = data.cost;
         this.dps = data.dps;
         this.level = data.level;
-        this.name = name.type;
+        this.name = data.name;
 
-        // Damage every .1 second
-        this.damageInterval = 0.1;
-        this.timer = setInterval(this.clickTarget.bind(this), this.damageInterval * 1000);
+        // Damage every .5 seconds
```
EthanVieira, line 38: Isn't this still every .1 seconds?

src/items/equipment.js
```diff
@@ -85,4 +100,15 @@ export default class Equipment extends Item {
         super.highlightItem();
         console.log("use", this.name);
     }
+
+    // TODO: be able to have multiple different required levels for different skills
+    checkRequiredLevel() {
+        let skill = this.skill.toLowerCase();
+        let level = calcLevel(this.scene.characterData.skills[skill]);
+        if (level >= this.requiredLevel) {
```
EthanVieira, line 108: return (level >= this.requiredLevel);

src/items/item.js
```diff
@@ -140,36 +141,40 @@ export class Item extends ClickableObject {
...
+                let visualNum = "0";
+                let fillColor = "";
+
+                // Set format/color based on the amount
+                switch (true) {
+                    case num < 1000:
+                        visualNum = num;
+                        fillColor = "orange";
+                        break;
+                    case num < 10000:
```
EthanVieira, line 160: White should be less than 10m and only have the 'k' if > 100k and the 'm' if > 10m. Also I don't think there is a space between the number and k/m

src/items/tools/knife.js
```diff
@@ -0,0 +1,53 @@
...
+    async craft(item) {
+        console.log("Combining", this.name, item.name);
+        let className = "";
+        let numRequiredItems = 0;
+        let xpGiven = 0;
+        switch (item.name) {
+            case "Logs":
+                className = "NormalShortbow";
+                numRequiredItems = 50;
+                xpGiven = 50;
```
EthanVieira, line 28: This seems a little light given it requires 50 logs. Maybe have it at 5xp per log?

src/items/item.js
EthanVieira: Looks like this also shows the decimal part (1.1k instead of 1100). If it has the k/m cutoff it should not show the decimals. I think the total amount of the stack should be visible via examine

src/items/tools/knife.js
```diff
@@ -0,0 +1,53 @@
...
+                break;
+            case "Oak Logs":
+                className = "OakShortbow";
+                numRequiredItems = 50;
+                xpGiven = 100;
```
EthanVieira, line 33: and then maybe 10 xp per log here

src/ui/dashboard.js
```diff
@@ -354,15 +289,8 @@ export class DashboardScene extends Phaser.Scene {
...
-            this.clan.scrollWindow = new ScrollWindow("clans");
-            this.scene.add(
-                "scroll-window",
-                this.clan.scrollWindow,
-                true,
-                this.characterData
-            );
-        }
+        this.clan.scrollWindow = new ScrollWindow("clans");
+        this.scene.add("scroll-window", this.clan.scrollWindow, true, this.characterData);
```
EthanVieira: I don't think this scroll window works - not sure if you've fixed in the future or not

src/ui/skill-descriptions.js
```diff
@@ -0,0 +1,48 @@
...
+            description += "Equip stronger magic weapons.";
+            body += "Lv 1: Equip staff\n";
+            body += "Lv 5: Equip magic staff\n";
+            body += "Lv 10: Equip staff of air\n";
+            body += "Lv 15: Equip staff of earth\n";
+            body += "Lv 20: Equip staff of water\n";
+            body += "Lv 25: Equip staff of fire\n";
+            break;
+        case "fletching":
+            description = "Create ranged weapons using a knife\nand logs";
```
EthanVieira, line 31: Hmm might wanna make these auto wrap. Fine to leave for now

src/ui/skills.js
```diff
@@ -0,0 +1,282 @@
...
+            index = 0;
+
+        // Add level text for skills
+        for (let skill in this.saveData) {
+            let row = index % 3;
+            let column = Math.floor(index / 3);
+            let x = startX + row * this.skillWidth;
+            let y = startY + column * this.skillHeight;
+
+            this.skillText[skill] = scene.add
```
EthanVieira, line 44: I think these are slightly to the right of the actual skill boxes - or at least it seems so

src/default-data.js
```diff
@@ -25,12 +26,28 @@ export const defaultData = {
...
         magic: 0,
-        health: 10,
+        fletching: 0,
         woodcutting: 0,
-        mining: 0,
+        runecrafting: 0,
+        slayer: 0,
+        farming: 0,
+        construction: 0,
+        trapping: 0,
```
EthanVieira, line 50: hunter :)

EthanVieira (changes requested):

src/items/item.js
EthanVieira: Should we add this to the fonts constant?

src/auto-clickers/auto-clicker.js
CurtisGreen: Yes, already fixed here https://github.com/CurtisGreen/rs-clicker/pull/1/files#diff-7c56f7094d1b1ea0a9983f199ae60ed86d21db156181649b9e10b01183a0f72b

src/items/equipment.js
CurtisGreen: Already fixed here: https://github.com/CurtisGreen/rs-clicker/pull/1/files#diff-74ef585e3bac3a0d35b3f94e8d160d27c57fc7f6e685ddf8a014028eb9e122c2

src/items/item.js
CurtisGreen: Adding it to the fonts constant would be good.

src/items/tools/knife.js
CurtisGreen: I think it should be increased but 5x is too much since you can chop logs way faster than in the normal game. Maybe 2x?

src/items/tools/knife.js
CurtisGreen: Same as above

src/ui/dashboard.js
CurtisGreen: I _think_ I made it better but it's still finicky sometimes

src/items/tools/knife.js
EthanVieira: I'm thinking everything in rs clicker will be much faster than the real game. Might as well keep the xp ratios for logs from wc to fletching the same

src/items/item.js
CurtisGreen: Lemme double check on this:

- 0-999: orange, show full # 
- 1000 - 99,999: white, show full # 
- 100k - 9999k: white, truncate and show k (this seems weird)
- 10m and above: green

src/items/tools/knife.js
CurtisGreen: Created a draft PR here: https://github.com/CurtisGreen/rs-clicker/pull/7. I'll make the text color changes when you give me confirmation on the format

src/items/item.js
EthanVieira: 0-99999: orange, show full # 
100k-9999k: white, truncate and show k
10m- 2147483647 (2147m, max signed 32bit): green, truncate and show m

EthanVieira (approved):

PR #46, opened 2021-03-13
All unlocked quests now show in the quest tab

EthanVieira:
image


CurtisGreen:
### Code
I know you copied the syntax from how I did `Clan`'s `scrollWindow` but I think it would be nicer if it could be encapsulated within the class. There might've been a good reason I did it like this but I think worst case you'll need to just add 1 or 2 accessor methods in QuestList and Clan to make it work.

### Visuals
1. Scroll bar is 1 pixel too far up. It might be 1 pixel too much to the left too but I can't tell.
1. Probably for future PRs but:
    a. Resources should probably be moved out from `enemiesKilled` so they aren't shown here (need to show them somewhere else though)
    b. Dark Wizard is too wide so the 'd' gets cut off
    c. It would be nice if the enemy names line up. Maybe this can be fixed by doing "01/10" instead of "1/10" so they're always 5 characters wide. Or maybe just put a space in front of single digits
    d. We need some visual notification that a quest has been completed. Either chat pops up, it switches to the quest tab, or the quest tab turns red or something
    e. Then we also need something that shows their newest unlock on the map


### Trello tasks
I added two new trello tasks for things I noticed that were important.
1. Visual notification a quest has been completed
2. Crafting status message should be printed to in game chat log (EX: Need X logs to craft that, or You created a bow, etc.)

src/constants/constants.js
```diff
@@ -16,6 +16,13 @@ export const CONSTANTS = {
         VARROCK: "VARROCK",
         BARBARIAN_VILLAGE: "BARBARIAN_VILLAGE",
     },
+    PREREQUISITES: {
```
CurtisGreen, line 19: I think this is more confusing than inputting the scene constant itself. Before I looked at constants I thought you duplicated them for some reason

CurtisGreen (commented):
### Code
I know you copied the syntax from how I did `Clan`'s `scrollWindow` but I think it would be nicer if it could be encapsulated within the class. There might've been a good reason I did it like this but I think worst case you'll need to just add 1 or 2 accessor methods in QuestList and Clan to make it work.

### Visuals
1. Scroll bar is 1 pixel too far up. It might be 1 pixel too much to the left too but I can't tell.
1. Probably for future PRs but:
    a. Resources should probably be moved out from `enemiesKilled` so they aren't shown here (need to show them somewhere else though)
    b. Dark Wizard is too wide so the 'd' gets cut off
    c. It would be nice if the enemy names line up. Maybe this can be fixed by doing "01/10" instead of "1/10" so they're always 5 characters wide. Or maybe just put a space in front of single digits
    d. We need some visual notification that a quest has been completed. Either chat pops up, it switches to the quest tab, or the quest tab turns red or something
    e. Then we also need something that shows their newest unlock on the map


### Trello tasks
I added two new trello tasks for things I noticed that were important.
1. Visual notification a quest has been completed
2. Crafting status message should be printed to in game chat log (EX: Need X logs to craft that, or You created a bow, etc.)

## Approval Requirements
1. Fix scroll bar pixel offset
2. See if scroll window can be initialized within `QuestList`
3. Convince me that the PREREQUISITES constants are better (or think of something even better)

src/ui/dashboard.js
```diff
@@ -228,18 +230,25 @@ export class DashboardScene extends Phaser.Scene {
...
+            height: 213,
+            numColumns: 1,
+            padding: 10,
         });
+        this.scene.add(this.quests.scrollWindow.name, this.quests.scrollWindow, true);
+        this.quests.scrollWindow.refresh();
 
-        // Set and hide quests on startup
-        this.updateKillQuestText();
+        this.quests.list = new QuestList(this, this.quests.scrollWindow);
```
CurtisGreen: Try initializing the scroll window in QuestList instead of passing it in. If it works and looks nice we should do it for Clan as well

src/constants/constants.js
EthanVieira: I like it because for all the places prereqs may be used we only have to think about them once when making this. Helps with code readability as well - a good example in this PR is [here](https://github.com/EthanVieira/rs-clicker/blob/a1d68f266fa8b01f3b8e9daaf3f61418363b7782/src/cookie-io.js#L182). If i were to try and replace that with scene constants it would be more code and more confusing. Also, if there is ever a level with multiple prereqs then it will be very helpful to have them combined in one place.

src/ui/dashboard.js
CurtisGreen: Okay I can see how it’s useful there. I guess for multiple prerequisites we would change it to a list and loop through them?

EthanVieira: I've done the deed but the scrollbars don't seem to be working... Just tested them on master and they don't seem to be working, so I guess something slipped through.

EthanVieira: > Okay I can see how it’s useful there. I guess for multiple prerequisites we would change it to a list and loop through them?

Yeah most likely we will need to handle all levels as having the potential for multiple prereqs and loop though them

CurtisGreen: > I've done the deed but the scrollbars don't seem to be working... Just tested them on master and they don't seem to be working, so I guess something slipped through.

Chat log and shop seem to be working well. I'm not sure why the clan window isn't; this scroll bar has given me so much pain lol. When you have a chance you can try your own rendition of it, I feel like I've rewritten it 3 times now

CurtisGreen (approved):
- [x] Fixed pixel offset
- [x] Moved scroll windows
- [x] Convinced me on the merits of prerequisite constant

PR #47, opened 2021-03-15
More quests

EthanVieira:
* There are now quest tiers for each enemy that are unlocked after completing the previous tier
* A level is complete when all tier 1 quests are completed for that level
* Quest text is now green when complete and yellow when incomplete
* Resource quests are not functional, but still have quest amounts

Screenshot 2021-03-14 220241
Screenshot 2021-03-14 220322
Screenshot 2021-03-14 220401

### DIFF
--- a/src/cookie-io.js
+++ b/src/cookie-io.js
@@ -190,12 +190,13 @@ class CharacterData {
         }
     }
 
-    incQuestTier(scene, name) {
-        this.characterData.levels[scene].questTier[name]++;
-    }
-
-    getQuestTier(scene, name) {
-        return this.characterData.levels[scene].questTier[name];
+    calcQuestTier(numKilled, tiers) {
+        for (const [index, tier] of tiers.slice(0, -1).reverse().entries()) {
+            if (numKilled >= tier) {
+                return tiers.length - index;
+            }
+        }
+        return 1;
     }
 
     getUnlockedLevels() {
--- a/src/cookie-io.js
+++ b/src/cookie-io.js
@@ -174,14 +174,14 @@ class CharacterData {
         }
     }
     setQuestCompleted(scene) {
-        if (this.checkScene(scene.currentLevel)) {
-            this.characterData.levels[scene.currentLevel].questCompleted = true;
+        if (this.checkScene(scene)) {
+            this.characterData.levels[scene].questCompleted = true;
 
             // Unlock levels
             for (var level in this.characterData.levels) {
                 if (
                     !this.characterData.levels[level].unlocked &&
-                    CONSTANTS.PREREQUISITES[level] == scene.currentLevel
+                    CONSTANTS.PREREQUISITES[level] == scene
                 ) {
                     this.characterData.levels[level].unlocked = true;
                     console.log("unlocking %s", level);
--- a/src/cookie-io.js
+++ b/src/cookie-io.js
@@ -174,18 +174,28 @@ class CharacterData {
         }
     }
     setQuestCompleted(scene) {
-        if (this.checkScene(scene)) {
-            this.characterData.levels[scene].questCompleted = true;
+        if (this.checkScene(scene.currentLevel)) {
+            this.characterData.levels[scene.currentLevel].questCompleted = true;
+
+            // Unlock levels
             for (var level in this.characterData.levels) {
                 if (
                     !this.characterData.levels[level].unlocked &&
-                    CONSTANTS.PREREQUISITES[level] == scene
+                    CONSTANTS.PREREQUISITES[level] == scene.currentLevel
                 ) {
                     this.characterData.levels[level].unlocked = true;
                     console.log("unlocking %s", level);
                 }
             }
         }
+    }
+
+    incQuestTier(scene, name) {
+        this.characterData.levels[scene].questTier[name]++;
+    }
+
+    getQuestTier(scene, name) {
+        return this.characterData.levels[scene].questTier[name];
     }
 
     getUnlockedLevels() {
--- a/src/default-data.js
+++ b/src/default-data.js
@@ -54,6 +54,9 @@ export const defaultData = {
         TUTORIAL_ISLAND: {
             questCompleted: false,
             unlocked: true,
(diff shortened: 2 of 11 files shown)


src/scenes/level.js
```diff
@@ -41,14 +42,11 @@ export class LevelScene extends Phaser.Scene {
...
         this.targetMetaData = data.targets;
+        this.questAmounts = data.questAmounts;
 
         // Store current level to return to after leaving shop
         this.currentLevel = data.key;
         this.levelType = data.levelType;
-
-        if (this.levelType == CONSTANTS.LEVEL_TYPE.ENEMY) {
-            this.killQuest = data.killQuest;
-        }
```
CurtisGreen, line 51: Nice, glad this got cleaned up

src/scenes/level.js
```diff
@@ -148,30 +146,55 @@ export class LevelScene extends Phaser.Scene {
 
     enemyKilled(name) {
         // Update kill quest score
-        if (characterData.getEnemiesKilled(this.currentLevel, name) < this.killQuest) {
+        if (
+            characterData.getEnemiesKilled(this.currentLevel, name) <
+            this.questAmounts[name][this.questAmounts[name].length - 1]
+        ) {
```
CurtisGreen: I think pulling out these conditions into two variables would help the readability here. Took me a bit to see what you were doing.

Maybe:

```js
const enemiesKilled = characterData.getEnemiesKilled(this.currentLevel, name);
const maxKillCount = this.questAmounts[name][ this.questAmounts[name].length - 1 ];
if (enemiesKilled < maxKillCount) {

```

src/cookie-io.js
```diff
@@ -174,12 +174,14 @@ class CharacterData {
...
-            this.characterData.levels[scene].questCompleted = true;
+        if (this.checkScene(scene.currentLevel)) {
+            this.characterData.levels[scene.currentLevel].questCompleted = true;
+
+            // Unlock levels
             for (var level in this.characterData.levels) {
                 if (
                     !this.characterData.levels[level].unlocked &&
-                    CONSTANTS.PREREQUISITES[level] == scene
+                    CONSTANTS.PREREQUISITES[level] == scene.currentLevel
```
CurtisGreen: Why pass in the whole scene object just to use the current level name like before? If you think `scene` is confusing because we usually use it to mean the object not the string, I think renaming the input to `sceneName` would be better.

src/default-data.js
```diff
@@ -54,13 +54,20 @@ export const defaultData = {
         TUTORIAL_ISLAND: {
             questCompleted: false,
             unlocked: true,
+            questTier: {
+                giantRat: 1,
+            },
```
CurtisGreen: We have access to current enemies killed and the array of quest tiers, what benefit is there in storing the current tier?

src/scenes/level.js
```diff
@@ -148,30 +146,55 @@ export class LevelScene extends Phaser.Scene {
...
+                let questCompleted = true;
+                this.targets.forEach((enemy, index) => {
+                    // Check for level completion.
+                    // A level is considered complete when
+                    // all of the tier 1 (index 0) quests are complete.
+                    if (
+                        characterData.getEnemiesKilled(this.currentLevel, enemy.varName) <
+                        this.questAmounts[enemy.varName][0]
+                    ) {
+                        questCompleted = false;
```
CurtisGreen: Looking at this now I'm not sure if this condition is necessary, quest completed should default to false.

Yeah `questCompleted` isn't needed now that we have `characterData.setQuestCompleted(this.currentLevel);`

CurtisGreen (changes requested):
Couple of things I commented on to clean up or clarify. Servers are borked at work so I can review but can't test lol

src/scenes/level.js
EthanVieira: Not sure I understand what you mean here. I think we do still need this because we need to set it to fail if any of the tier 1 quests are not completed.

src/default-data.js
EthanVieira: I considered adding this to the enemies killed or somehow just using the enemies killed amount but it resulted in much more unreadable code. I figured that this was a cleaner way to do it albeit somewhat duplicated efforts.

src/cookie-io.js
EthanVieira: At one point I was using another attribute of scene and I needed to pass in the object. [Reverted](https://github.com/EthanVieira/rs-clicker/pull/47/commits/0d1df380ee541e72d588204b3ac45a7aa2cea81e])!

src/scenes/level.js
EthanVieira: [Updated](https://github.com/EthanVieira/rs-clicker/pull/47/commits/0d1df380ee541e72d588204b3ac45a7aa2cea81e)!

src/scenes/level.js
CurtisGreen: I guess it's related to the comment above on default-data. I was thinking just looing at `enemiesKilled` and calculating the tier would be simple and more readable. But you're saying it was less readable somehow?

```js
function calcTier(numKilled, tiers) {
    tiers.slice.reverse.forEach((tier, index) => {
        if (numKilled >= tier) {
            return index;
        }
    }
    return 0;
}
```

If you really think this is worse than adding/using the getters and setters and the `questTier`s then I'll believe you since you said you tried both

CurtisGreen: Btw you can link github PRs to trello tasks and it'll show their current status which is neat. Just copy the PR link, click to expand the story and then press ctrl+v.

Btw I'm using github's task system for clash.tips and it's okay but it's not as good as trello yet.

EthanVieira: Ah I like the calcTier method. That is much better.

EthanVieira: [Updated](https://github.com/EthanVieira/rs-clicker/pull/47/commits/fdcb36749bb7a9e1cf8f8fffdbe35ab9ec40af13)!

CurtisGreen (approved):

PR #48, opened 2021-03-20
Chat Log Notifications

CurtisGreen:
# Chat log notifications
image
image

### Features
- Crafting now prints info (success / fail) to the chat
- Level ups now print to chat
- Quest unlocks now print to chat
- Chat button turns blue when you write text to the chat log and turns back normal when you click on it

### Fixes
- Fixed a bug where pressing "Log" then opening the map or exiting would leave the log window on top

### Adjustments
- Created a new function called `writeText` in chat window that does a one shot print to chat
- The other function `writeStrings` pushes the strings and formatting but it doesn't render until you refresh, show, etc. for other cases where you want to write a bunch before rendering

### New bugs found
- Varrock guard / dark mages don't show up in the quest window


src/cookie-io.js
```diff
@@ -139,14 +139,21 @@ class CharacterData {
...
+            const prevLevel = Utilities.calcLevel(this.characterData.skills[skill]);
             this.characterData.skills[skill] += xp;
-            const curLevel = calcLevel(this.characterData.skills[skill]);
+            const curLevel = Utilities.calcLevel(this.characterData.skills[skill]);
 
             // Play level up sfx
             if (curLevel > prevLevel) {
                 const audioScene = this.getScene(CONSTANTS.SCENES.AUDIO);
                 audioScene.playSfx(skill + "-level-up");
+
```
EthanVieira, line 150: Can we add a console.log here as well? Seems like good debug info to have

src/cookie-io.js
```diff
@@ -139,14 +139,21 @@ class CharacterData {
...
+                // Write level up text to chat
+                const chatScene = this.getScene(CONSTANTS.SCENES.CHAT);
```
EthanVieira, line 152: Is there a way we can encapsulate the getScene in the writeText method? would be nice to just have a single method to write to chat

src/items/tools/knife.js
```diff
@@ -34,22 +36,32 @@ export default class Knife extends Tool {
                 xpGiven = 500;
                 break;
             default:
-                console.log("Not a valid crafting combination.");
+                outputString = "Not a valid crafting combination.";
                 break;
         }
 
+        // Craft item if possible
         if (className != "" && item.numItems >= numRequiredItems) {
```
EthanVieira: Thoughts on putting crafting logic into a Tool method? Thinking a good amount of this logic might be duplicated with other tools

src/ui/chat-window.js
```diff
@@ -90,7 +90,18 @@ export class ChatScene extends Phaser.Scene {
         this.show(false);
     }
 
-    // Create a row of text on the scroll window
+    // Takes in text and refreshes the display
+    writeText(text, format = FONTS.ITEM_HEADER) {
+        this.show();
```
EthanVieira: I'm not sure I like force showing the log on every update... but I think the reason is that it hides part of the monster hitbox. I think ultimately it should show by default, but if the user hides the log then it shouldn't pop up on every update.

Also should probably just move the monsters up a bit so that having the log out doesn't make it annoying

src/utilities.js
```diff
@@ -47,3 +47,19 @@ export function prettyPrintCamelCase(str) {
...
     return resultStr;
 }
+
+// Transforms VARROCK_MINE -> Varrock Mine
+export function prettyPrintConstant(str) {
+    let resultStr = "";
+    const spaceStr = str.replace("_", " ");
+    for (let i = 0; i < spaceStr.length; i++) {
+        if (i == 0 || spaceStr.charAt(i-1) == " ") {
+            resultStr += spaceStr[i].toUpperCase();
```
EthanVieira: Is there a difference btwn using .charAt() or [i]? If not, let's use the same to be consistent

EthanVieira (changes requested):
Left a few comments but overall pretty noice. Honestly I think this completes the trello task I had in progress as well so I'm gonna move that over and delete the branch I created for it.

src/ui/chat-window.js
EthanVieira: Also if the log is hiding and there is an update written to it, the show/hide log button gets highlighted. That would be ideal here as well

src/cookie-io.js
CurtisGreen: You have to get the object to call the function somehow. We could add a global function inside the same file to simplify it but I'm not sure if I like that or not

src/items/tools/knife.js
CurtisGreen: Yeah I was thinking about that too

src/ui/chat-window.js
CurtisGreen: Alternatively if you click on the enemy it could close the log

src/utilities.js
CurtisGreen: I think the only difference is charAt returns "" if you go out of range whereas [] returns undefined. This was just a mistake though

src/items/tools/knife.js
CurtisGreen: Change here: https://github.com/EthanVieira/rs-clicker/pull/48/commits/28f8b798b6dff330adcda83d064e7b3c292dd074

src/utilities.js
CurtisGreen: Changed here https://github.com/EthanVieira/rs-clicker/pull/48/commits/2508de4ec12ce492b7f36b20ba0644379db853d7

src/ui/chat-window.js
EthanVieira: I think the only thing that should show/hide the log is the one button in the bottom left.

src/utilities.js
EthanVieira: There's one more a couple lines under 😄

src/cookie-io.js
CurtisGreen: Just put the console log inside the writeText function here: https://github.com/EthanVieira/rs-clicker/pull/48/commits/1ef3458f169d96c74b1bf83b07d6ef5eed9392f6

src/utilities.js
CurtisGreen: Alright https://github.com/EthanVieira/rs-clicker/pull/48/commits/a59b4519bcb36cfae50a00740189bcc6ee36cd00

src/ui/chat-window.js
CurtisGreen: Alright I'll probably look into making some sort of indication that it's been updated without popping it up

CurtisGreen: Done, see updated screenshot in the description

src/ui/chat-window.js
```diff
@@ -89,15 +91,15 @@ export class ChatScene extends Phaser.Scene {
...
+    // Takes in text, refreshes display, and shows notification button
     writeText(text, format = FONTS.ITEM_HEADER) {
-        this.show();
-        this.chatWindow.visible = true;
         this.writeStrings(
             { x: 0, text, format },
         );
         this.scrollWindow.refresh();
         this.scrollWindow.scrollToBottom();
+        this.show(false);
```
EthanVieira: Small thing here - if it is currently showing then it shouldn't close. Only keep closed if it is already closed

CurtisGreen: Alright check again

EthanVieira (approved):

PR #49, opened 2021-03-27
Sell when dropping

EthanVieira:
* Drop in right click menu is now sell
* Added currency item type which gets loaded in with a list of images such that the stack can change as it increases
* Gold is now stored in the inventory as an item (start with 25 like in RS)
* Stats updated to track total earned gold
* Shop no longer displays gold. Whenever we update it to be a window pop-up, you will be able to see the gold in your inventory. Currently you cannot see current gold while in the shop.
* Added smaller coin stacks 1-5


CurtisGreen (approved):
LGTM 😎

PR #51, opened 2021-03-30
log required level on equip failure

EthanVieira:
image

### DIFF
--- a/src/items/equipment.js
+++ b/src/items/equipment.js
@@ -1,6 +1,12 @@
 import { Item } from "./item.js";
-import { CONSTANTS, FONTS, OBJECT_TYPES, EQUIPMENT } from "../constants/constants.js";
-import { calcLevel, getItemClass, capitalize, aOrAn } from "../utilities.js";
+import { CONSTANTS, OBJECT_TYPES } from "../constants/constants.js";
+import {
+    calcLevel,
+    getItemClass,
+    capitalize,
+    aOrAn,
+    getRequiredCombatSkill,
+} from "../utilities.js";
 import { characterData } from "../cookie-io.js";
 
 export default class Equipment extends Item {
@@ -73,7 +79,7 @@ export default class Equipment extends Item {
                 equippedItem.scene.equipment.obj.equipItem(equippedItem);
             } else {
                 console.log("Not high enough level to equip that.");
-                let skillText = this.getRequiredCombatSkill();
+                let skillText = getRequiredCombatSkill(this.skill);
                 this.scene.scene
                     .get(CONSTANTS.SCENES.CHAT)
                     .writeText(
@@ -115,23 +121,8 @@ export default class Equipment extends Item {
 
     checkRequiredLevel() {
         return (
-            calcLevel(characterData.getSkillXp(this.getRequiredCombatSkill())) >=
+            calcLevel(characterData.getSkillXp(getRequiredCombatSkill(this.skill))) >=
             this.requiredLevel
         );
-    }
-
-    getRequiredCombatSkill() {
-        let skill = "";
-        switch (this.skill) {
-            case EQUIPMENT.WEAPON_TYPES.MELEE:
-                skill = "attack";
-                break;
-            case EQUIPMENT.WEAPON_TYPES.RANGED:
-                skill = "ranged";
-                break;
-            case EQUIPMENT.WEAPON_TYPES.MAGIC:
-                skill = "magic";
-        }
-        return skill;
     }
 }
--- a/src/items/equipment.js
+++ b/src/items/equipment.js
@@ -1,6 +1,6 @@
 import { Item } from "./item.js";
-import { OBJECT_TYPES } from "../constants/constants.js";
-import { calcLevel, getItemClass } from "../utilities.js";
+import { CONSTANTS, FONTS, OBJECT_TYPES, EQUIPMENT } from "../constants/constants.js";
+import { calcLevel, getItemClass, capitalize, aOrAn } from "../utilities.js";
 import { characterData } from "../cookie-io.js";
 
 export default class Equipment extends Item {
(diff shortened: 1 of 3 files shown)


src/items/equipment.js
```diff
@@ -101,14 +113,25 @@ export default class Equipment extends Item {
...
-        }
+        return (
+            calcLevel(characterData.getSkillXp(this.getRequiredCombatSkill())) >=
+            this.requiredLevel
+        );
+    }
 
-        let level = calcLevel(characterData.getSkillXp(skill));
-        return level >= this.requiredLevel;
+    getRequiredCombatSkill() {
```
CurtisGreen: Can you use this function to clean up some of the logic in https://github.com/EthanVieira/rs-clicker/blob/master/src/targets/enemy.js? Particularly ` getDamageLevel()` and `increaseXp()`

Or do you think that's out of the scope of this PR?

src/utilities.js
```diff
@@ -81,3 +81,13 @@ export function getGoldStackType(goldAmount) {
             return "10k";
     }
 }
+
+export function capitalize(str) {
+    return str.charAt(0).toUpperCase() + str.slice(1);
+}
+
+export function aOrAn(followingWord) {
+    return ["a", "e", "i", "o", "u"].includes(followingWord.charAt(0).toLowerCase())
```
CurtisGreen, line 92: 🧠

CurtisGreen (changes requested):

src/items/equipment.js
CurtisGreen: Also maybe put a print if the default condition happens because under normal circumstances it should never happen

EthanVieira: Yessir, [updated](https://github.com/EthanVieira/rs-clicker/pull/51/commits/82e4e87e3f5a42b9ade56c8027b12ac6b0bf5ae0)!

EthanVieira: and [here](https://github.com/EthanVieira/rs-clicker/pull/51/commits/4511f9c2fe3b1749e72e4c7deae355319819377a)

src/targets/enemy.js
```diff
@@ -253,22 +252,11 @@ export class Enemy extends Target {
...
-                    return calcLevel(characterData.getSkillXp("magic"));
-                    break;
-                case EQUIPMENT.WEAPON_TYPES.RANGED:
-                    return calcLevel(characterData.getSkillXp("ranged"));
-                    break;
-                case EQUIPMENT.WEAPON_TYPES.MELEE:
-                    return calcLevel(characterData.getSkillXp("attack"));
-                    break;
-            }
-        } else {
```
CurtisGreen, line 269: This else is needed otherwise unarmed does 0 damage and gives 0 xp

src/targets/enemy.js
```diff
@@ -277,23 +265,10 @@ export class Enemy extends Target {
...
-                    characterData.addSkillXp("magic", xpIncrease);
-                    break;
-                case EQUIPMENT.WEAPON_TYPES.RANGED:
-                    characterData.addSkillXp("ranged", xpIncrease);
-                    break;
-                case EQUIPMENT.WEAPON_TYPES.MELEE:
-                    characterData.addSkillXp("attack", xpIncrease);
-                    break;
-            }
-        } else {
```
CurtisGreen, line 294: Same here

src/utilities.js
```diff
@@ -91,3 +93,18 @@ export function aOrAn(followingWord) {
         ? "an"
         : "a";
 }
+
+export function getRequiredCombatSkill(skillConstant) {
```
CurtisGreen, line 97: I think keeping it as a method of the equipment class was fine since you could do: `this.equipment.obj.equipment.WEAPON.getRequiredCombatSkill()` but it doesn't really make much of a difference I guess

EthanVieira: Ehhhh I prefer utilities. Doesn't really need to be attached to equipment

src/targets/enemy.js
EthanVieira: [Fixed](https://github.com/EthanVieira/rs-clicker/pull/51/commits/3cc7bbd09b9aa339abf9b002707b19174241c7df)!

CurtisGreen: image
Lol

This can be fixed by keeping that `if (object exists) {} else { "unarmed"}` check. That way for the normal one you can still have error checking on it. On that note I think the `Object.entries(obj).length` might be a good utility function to add as well. Can be called `exists(obj)` or objExists(obj)` or something

CurtisGreen: The screenshot is when you click while unarmed

src/utilities.js
```diff
@@ -107,7 +107,10 @@ export function getRequiredCombatSkill(skillConstant) {
             skill = "magic";
```
CurtisGreen: These can all be returned instead of setting "skill" then returning it since we're using a default now

src/utilities.js
```diff
@@ -107,7 +107,10 @@ export function getRequiredCombatSkill(skillConstant) {
...
             break;
         default:
-            console.log("Could not find the required combat skill.");
+            skill = "attack";
+            console.log(
+                "Could not find the required combat skill. Defaulting to unarmed."
+            );
```
CurtisGreen: Currently prints out every time you click while unarmed. Can be fixed by adding an optional parameter
```js
export function getRequiredCombatSkill(skillConstant, allowUnarmed = false) {
  //...
  default:
    if (!allowUnarmed ) {
      console.log('Could not find the required combat skill. Defaulting to attack.');
    }
    return attack
}
```

src/targets/enemy.js
EthanVieira: Decided to remove log. https://github.com/EthanVieira/rs-clicker/pull/51/commits/b73bf2b6aa7cc726e0d5e6d0c3054dfbdc5bddf7

src/utilities.js
EthanVieira: [Updated](https://github.com/EthanVieira/rs-clicker/pull/51/commits/1e14e04e305e70c3386791a2c9308c511e85bc7a).

CurtisGreen (approved):

PR #54, opened 2021-04-10
Dashboard cleanup

CurtisGreen:
Moved panel, button, and show() logic from Dashboard into the individual classes (Inventory, Equipment, etc.) so dashboard is significantly simpler.

Also moved the panel classes into a folder called `panels` .

The general structure of code is pretty similar between the panels so I think there should be a way to a way to further reduce code duplication through a panel parent class but due to some little differences I refrained from doing that this time.


EthanVieira (changes requested):
We should probably have a dashboard panel class or something that all these inherit from to save duplicate code but w/e I won't make you do it.

One thing that I think I just missed on the previous PR is that when you have all songs unlocked, you get the congrats message every time you switch areas. Should only happen once.

CurtisGreen:
> We should probably have a dashboard panel class or something that all these inherit from to save duplicate code but w/e I won't make you do it.
> 
> One thing that I think I just missed on the previous PR is that when you have all songs unlocked, you get the congrats message every time you switch areas. Should only happen once.

I wanted to make the class but there are a variable number of objects to show/hide in each one so I’m not sure the best way to do it. Maybe put them in a dictionary? 

Also I think I fixed that bug with this PR, try it out

EthanVieira:
> I wanted to make the class but there are a variable number of objects to show/hide in each one so I’m not sure the best way to do it. Maybe put them in a dictionary?
> 
> Also I think I fixed that bug with this PR, try it out

I noticed the song unlock log when testing this branch so I don't think it is fixed.

CurtisGreen:
> I noticed the song unlock log when testing this branch so I don't think it is fixed.

Alright I think I fixed it. Also I added a utilities function `runOnLoad()` that should clean up some scenarios where functions are called before scenes have been created. I had some issues with autoclickers/the update function being used to call scenes immediately

EthanVieira:
> 
> Alright I think I fixed it. Also I added a utilities function `runOnLoad()` that should clean up some scenarios where functions are called before scenes have been created. I had some issues with autoclickers/the update function being used to call scenes immediately

If you have the clan panel open and then go to shop and buy autoclickers, the "bot" text will be shown on any dashboard panel. Not sure what caused it but don't think that was there before?

CurtisGreen:
> If you have the clan panel open and then go to shop and buy autoclickers, the "bot" text will be shown on any dashboard panel. Not sure what caused it but don't think that was there before?

I think I know what caused this, lemme try

CurtisGreen:
Fixed

EthanVieira (approved):

PR #55, opened 2021-04-16
Right click menu

EthanVieira:
The right click menu now: 
* closes when clicked anywhere on it
* is correctly sized based on the number of options
* never gets permanently stuck (afaik)
* spawns under your cursor (unless on bottom edge)
* will reposition to not spawn off screen

Known bugs:
* right click + hold for ~.5 seconds and then move off of menu while still holding causes the menu to not disappear unless re-interacted with
* right click + super fast left click will do the same thing as above
* spawns underneath chat log depth
* text/image is lower quality on some items/areas of screen


CurtisGreen (approved):
Simpler than I expected it to be, nice job

PR #56, opened 2021-04-17
Selling updates

EthanVieira:
* created a text prompt that is currently being used for selling x amount of an item


src/ui/chat-window.js
```diff
@@ -18,11 +18,75 @@ export class ChatScene extends Phaser.Scene {
...
+    isPromptOpen = false;
 
     constructor() {
         super({ key: CONSTANTS.SCENES.CHAT });
     }
 
+    // TODO: if we need more prompts we should reuse a lot of this
+    // Right now it is specifically for selling X amount of an item
+    prompt(promptText, item) {
+        let originallyVisible = this.visible;
```
CurtisGreen: Small thing, we should use const unless we plan on modifying the value. Not specifically for performance reasons but for ease of understanding the code

CurtisGreen (changes requested):
Looks nice but it doesn't work with numpad inputs as I think those have separate key IDs.

CurtisGreen:
I also think the sell X menu should close when you press escape

EthanVieira:
> Looks nice but it doesn't work with numpad inputs as I think those have separate key IDs.

Hmm the 96-105 range should be the numpad numbers

src/ui/chat-window.js
EthanVieira: Updated! https://github.com/EthanVieira/rs-clicker/pull/56/commits/ca5f638ee5466c01f3ba32b12dcd04aaceda9e03

src/ui/chat-window.js
```diff
@@ -18,11 +18,86 @@ export class ChatScene extends Phaser.Scene {
...
+                // arbitrary input length limit
+                if (promptInput.text.length < 15) {
+                    promptInput.text += event.key;
+                    promptInput.x -= 4;
+                }
+            }
+
+            // esc
+            if (event.keyCode == 27) {
+                promptWindow.destroy();
```
CurtisGreen: Since this is duplicate code as above maybe put it into its own function for show/hide prompt

src/ui/chat-window.js
```diff
@@ -18,11 +18,86 @@ export class ChatScene extends Phaser.Scene {
...
+
+                promptWindow.destroy();
+                prompt.destroy();
+                promptInput.destroy();
+                this.removeAllListeners();
+
+                item.sellX(inputValue);
+
+                this.scene.isPromptOpen = false;
+                this.scene.show(originallyVisible);
```
CurtisGreen: If something is sold, it should write to chat saying so

EthanVieira: Updated! https://github.com/EthanVieira/rs-clicker/pull/56/commits/93c33cfc10a7f52b57dad5438451f5d67196d089

src/ui/chat-window.js
EthanVieira: Added! https://github.com/EthanVieira/rs-clicker/pull/56/commits/a95ee710781df2120dcc8bca079e57c6cbd8f58e

CurtisGreen (approved):

PR #57, opened 2021-04-17
update resource levels

EthanVieira:
* now checks for both the required level of the tool (axe/pickaxe) you are using and the required level of the resource
* allow you to use a tool in your inventory to click the resource
* stats now show correctly on resource levels
* quests now update for resource levels

### DIFF
--- a/src/scenes/level.js
+++ b/src/scenes/level.js
@@ -1,5 +1,6 @@
 import { CONSTANTS, EQUIPMENT } from "../constants/constants.js";
 import { characterData } from "../cookie-io.js";
+import { getItemClass } from "../utilities.js";
 
 export class LevelScene extends Phaser.Scene {
     // General info that all levels should implement
@@ -22,6 +23,7 @@ export class LevelScene extends Phaser.Scene {
     targetMetaData = [];
     currentTargetIndex = 0;
     levelType = "";
+    resourceType = "";
     // For enemy levels
     // varName: amount
     // giantRat: 10
@@ -47,6 +49,7 @@ export class LevelScene extends Phaser.Scene {
         // Store current level to return to after leaving shop
         this.currentLevel = data.key;
         this.levelType = data.levelType;
+        this.resourceType = data.resourceType;
     }
 
     preload() {
@@ -197,7 +200,11 @@ export class LevelScene extends Phaser.Scene {
 
         // Get current weapon image
         let curWeapon = this.dashboard.equipment.equipment.WEAPON;
-        if (Object.keys(curWeapon).length) {
+
+        if (
+            Object.keys(curWeapon).length &&
+            this.levelType == CONSTANTS.LEVEL_TYPE.ENEMY
+        ) {
             // Set animation based on current weapon
             switch (curWeapon.skill) {
                 case EQUIPMENT.WEAPON_TYPES.MELEE:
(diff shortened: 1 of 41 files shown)


src/scenes/level.js
```diff
@@ -250,6 +257,47 @@ export class LevelScene extends Phaser.Scene {
...
+                default:
+                    console.log("Error: this resource type does not exist.");
+                    break;
+            }
+
+            if (toolEquipped) {
+                imageName = curWeapon.sprite.texture.key + "-model";
+            } else {
+                let itemObj = getItemClass(
+                    characterData.getInventory()[i].item,
```
CurtisGreen, line 293: Do you think we should add a function for this?

src/targets/resources/rock.js
```diff
@@ -9,10 +13,46 @@ export class Rock extends Resource {
...
+                miningLevel >=
+                getItemClass(characterData.getInventory()[i].item, this.scene)
+                    .requiredMiningLevel;
+        } else {
+            chat.writeText("This action requires a pickaxe.");
+            return false;
+        }
+
+        // Currently a small bug here that will not let you mine if the pickaxe you
+        // are unable to use is higher in the inventory than another pickaxe that you can use
```
CurtisGreen: Would it be hard to loop through and find all useable pickaxes and then take the one with the highest level?

src/targets/resources/tree.js
```diff
@@ -9,10 +13,45 @@ export class Tree extends Resource {
     constructor(data) {
         data.skill = "woodcutting";
         super(data);
+        this.requiredWCLevel = data.requiredWCLevel;
     }
 
     isClickable() {
         let curWeapon = this.scene.dashboard.equipment.equipment.WEAPON;
-        return curWeapon.item == "Axe";
+        let inventory = this.scene.dashboard.inventory;
```
CurtisGreen: I think this function should be generalized and moved into the resource class because it's shared between woodcutting and mining and if we add another resource later it will also require something similar. As far as I can tell the only thing you would need to do is pass in "Axe" and "woodcutting". You might also need to change`requiredWCLevel` into:
```
requiredLevels = {
    attack: 5,
    woodcutting: 6
}
```

Or whatever is easiest

src/ui/panels/inventory.js
```diff
@@ -89,6 +89,21 @@ export class Inventory {
         return -1;
     }
 
+    // Returns first instance of an item in inventory
+    // that contains the given keyword in its name
+    getKeywordInInventory(keyword) {
+        let playerItems = characterData.getInventory();
+        for (let i = 0; i < playerItems.length; i++) {
```
CurtisGreen, line 102: Something like
```js
// skill = "woodcutting"
playerItems = playerItems.filter((item) => item.item.includes(keyword)); // Not sure if we need the object check or not...
playerItems.sort((a, b) => b.requiredLevel[skill] - a.requiredLevel[skill]);
if (playerItems.length > 0) {
    return playerItems[0].index;
} else {
    return -1;
}
```

CurtisGreen (changes requested):

src/scenes/level.js
EthanVieira: This exact stuff should only be used for creating the animation so I'd say leave it

src/targets/resources/rock.js
EthanVieira: updated! https://github.com/EthanVieira/rs-clicker/pull/57/commits/8048b8819e2c3b3ec52cc513a0ede33b9f57569a

src/targets/resources/tree.js
EthanVieira: updated! https://github.com/EthanVieira/rs-clicker/pull/57/commits/8048b8819e2c3b3ec52cc513a0ede33b9f57569a

src/ui/panels/inventory.js
EthanVieira: updated! https://github.com/EthanVieira/rs-clicker/pull/57/commits/8048b8819e2c3b3ec52cc513a0ede33b9f57569a

src/ui/panels/inventory.js
```diff
@@ -42,7 +42,7 @@ export class Inventory {
             let item = playerItems[i];
             if (Object.keys(item).length) {
                 // Create item from name
-                let itemObj = await getItemClass(item.item, this.scene);
```
CurtisGreen: Looks like this one is required

CurtisGreen: nvm wrong one

src/ui/panels/equipment.js
```diff
@@ -56,7 +56,7 @@ export class Equipment {
 
         for (let i in playerEquipment) {
             if (Object.keys(playerEquipment[i]).length) {
-                let newEquipment = await getItemClass(playerEquipment[i], this.scene);
```
CurtisGreen: W/o this it unequips the item on scene change for some reason

CurtisGreen (changes requested):
image
Examine goes off the screen because of multiple required levels. Maybe make it like the stats where it's:
```
Required levels:    Attack  5  Mining  6
Accuracy Bonuses:   Stab    8  Slash   -2  Crush   6    Mag    0   Rng    0
...
```
It's not a deal breaker for me though since you can just scroll up

CurtisGreen (approved):

PR #58, opened 2021-04-18
Volume slider

CurtisGreen:
- Volume sliders are no longer buttons, instead they can be dragged and they snap to the closest position.
- Put slider backgrounds back in the panel picture itself so we don't have needless extra image loads + creates


EthanVieira (approved):
noice noice noice

PR #59, opened 2021-04-20
user can now type into chat

EthanVieira:
* also changed the input prompt character to be at the end of input instead of the beginning
image


src/ui/chat-window.js
```diff
@@ -382,7 +390,7 @@ export class ChatScene extends Phaser.Scene {
             // enter
             if (event.keyCode == 13) {
                 let inputValue = parseInt(
-                    promptInput.text.substr(1, promptInput.text.length)
+                    promptInput.text.substr(0, promptInput.text.length - 1)
```
CurtisGreen, line 393: Why can't you just do `promptInput.text`?

src/ui/chat-window.js
```diff
@@ -409,7 +415,10 @@ export class ChatScene extends Phaser.Scene {
             ) {
                 // arbitrary input length limit
                 if (promptInput.text.length < 15) {
-                    promptInput.text += event.key;
+                    promptInput.text =
+                        promptInput.text.substr(0, promptInput.text.length - 1) +
```
CurtisGreen, line 419: Same thing here

src/ui/chat-window.js
```diff
@@ -432,4 +441,78 @@ export class ChatScene extends Phaser.Scene {
         this.isPromptOpen = false;
         this.show(showChat);
     }
+
+    allowTyping() {
+        this.userMessage.visible = true;
+
+        // Don't create duplicate listeners if this is called twice
+        if (this.chatInput != null) {
```
CurtisGreen, line 449: Do you think we should replace our object instantiations and conditional checks with null? Might get rid of unnecessary `Object.keys(obj).length` checks, or w/e it is

src/ui/chat-window.js
```diff
@@ -432,4 +441,78 @@ export class ChatScene extends Phaser.Scene {
...
+            this.chatInput.removeAllListeners();
+        }
+
+        this.chatInput = this.input.keyboard.on("keydown", function (event) {
```
CurtisGreen: We should probably be consistent in using ES6 notation like:
```js
this.chatInput = this.input.keyboard.on("keydown", (event) => {
```

CurtisGreen (changes requested):
Overall looking good but I wanna double check some things. Also with this + the previous PR chat-window is starting to get a bit bulky. If you can think of ways to reduce duplicate code or split things up lmk

EthanVieira:
> Overall looking good but I wanna double check some things. Also with this + the previous PR chat-window is starting to get a bit bulky. If you can think of ways to reduce duplicate code or split things up lmk

This file should be more or less done at this point. I think the keyboard input schemes for prompt/chat should be separate and a lot of the other methods should probably stay here as well.

src/ui/chat-window.js
EthanVieira: The text also has the prompt character '*' at the end of it

src/ui/chat-window.js
EthanVieira: hmmm I'm not sure.. Don't know enough about JS to know which is better practice

src/ui/chat-window.js
EthanVieira: Yessir - updated! https://github.com/EthanVieira/rs-clicker/pull/59/commits/3a4179fa0536cef2fb2618102ede1b19bb1e7fc7

src/ui/chat-window.js
CurtisGreen: ah

EthanVieira:
> > Overall looking good but I wanna double check some things. Also with this + the previous PR chat-window is starting to get a bit bulky. If you can think of ways to reduce duplicate code or split things up lmk
> 
> This file should be more or less done at this point. I think the keyboard input schemes for prompt/chat should be separate and a lot of the other methods should probably stay here as well.

If we were to move something it should be the written presets like "writeEquipmentInfo". I imagine those could belong in their own file

src/ui/chat-window.js
CurtisGreen: Alright, I'll leave it as-is and maybe experiment next PR I do

src/ui/chat-window.js
```diff
@@ -409,14 +415,17 @@ export class ChatScene extends Phaser.Scene {
...
+                        event.key +
+                        "*";
                     promptInput.x -= 4;
                 }
             }
 
             // esc
             if (event.keyCode == 27) {
-                this.scene.destroyPrompt(promptObjs, keyboardInput, originallyVisible);
+                this.destroyPrompt(promptObjs, keyboardInput, originallyVisible);
```
CurtisGreen, line 428: I think I'd prefer that escape closes chat rather than just removes the current prompt. This applies to both this and the previous one. For this one just one escape closes chat. The other prompt would need 1 to close prompt and another to close chat if it's still open.

Pressing enter or space to open the chat log would be convenient too

EthanVieira: I decided to do half of what you said - https://github.com/EthanVieira/rs-clicker/pull/59/commits/524288079e5373beddba435a63033be79b5845a5

CurtisGreen (approved):

PR #60, opened 2021-04-24
Al kharid furnace

CurtisGreen:
image

### Additions
- Al Kharid furnace scene
- `Furnace` object that extends `ClickableObject` (instead of `Target`)
- BronzeBar
- Smithing level up SFX

### Fixes
- Error in `Inventory::getKeywordInInventory()` if you ran it with `mustBeUsable` = false.

### Notes
- Smelt ore by clicking on a desired ore and then clicking the furnace. Because clicking on the furnace does not deselect the ore you can spam it as usual.
- `Furnace` is implemented differently than other level targets. It didn't make sense to add a duplicate image overlaid on the background so it just checks for clicks in a given rectangle on the screen. You can pass in `{debug: true}` into the `Button` constructor to see an outline when you're trying to place it.


src/README.md
```diff
@@ -19,15 +20,28 @@
...
+| -------------- | --------------------------------------------------------------------------------------------------------------------------------------- |
+| `main.js`      | Sets up phaser and loads all of the scenes. Scene load order affects the z-index                                                        |
+| `load.js`      | Init cookies, load item/panel images, and dynamically import all item classes from `*-manifest.js` files and store them in a dictionary |
+| `cookie-io.js` | Contains getter/setters for cookie obj. The obj reference is passed between all scenes to get/set save data                             |
+| `main-menu.js` | Starts music from `scenes/audio.js`, lets you delete cookies, and start on the last level the user was on                               |
+| `level.js`     | Parent class for all levels. Creates the background, targets, and dashboard UI. Also checks quest completion and animates weapon swing  |
+| `dashboard.js` | Contains the visuals and objects for the right portion of the screen (inventory, skills, etc.)                                          |
+
+## Where to add content
```
EthanVieira, line 35: Not sure I see the full purpose of this section

src/targets/furnace.js
```diff
@@ -0,0 +1,98 @@
...
+
+    constructor(scene) {
+        super();
+        this.scene = scene;
+
+        const cameraWidth = scene.cameras.main.width;
+        const cameraHeight = scene.cameras.main.height;
+        const x = cameraWidth / 2 - 230;
+        const y = cameraHeight / 2 - 150;
+        const width = 230;
```
EthanVieira, line 22: Do we want to define width/height a few lines above so we can use them in the x/y definitions?

src/targets/furnace.js
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    async clickTarget() {
+        const inv = this.scene.dashboard.inventory;
+        const chat = this.scene.scene.get(CONSTANTS.SCENES.CHAT);
+
+        // See if ore is selected
+        const selectedIndex = inv.curSelectedItemIndex;
+        if (selectedIndex < 0) {
+            chat.writeText("Select an ore in your inventory first");
```
EthanVieira: Add a period here to be consistent

src/targets/furnace.js
```diff
@@ -0,0 +1,98 @@
...
+            case "Tin Ore":
+                this.smelt(
+                    ["CopperOre", "TinOre"],
+                    "BronzeBar",
+                    1,
+                    "You need at least 1 copper and 1 tin in your inventory."
+                );
+                break;
+            default:
+                chat.writeText("The furnace can only be used with an ore selected");
```
EthanVieira: Period here also

src/targets/furnace.js
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        const selectedItem = inv.inventory[selectedIndex];
+        switch (selectedItem.name) {
+            case "Copper Ore":
+            case "Tin Ore":
+                this.smelt(
+                    ["CopperOre", "TinOre"],
+                    "BronzeBar",
+                    1,
```
EthanVieira: RS gives 6.25 xp for bronze bars

EthanVieira: xp should also probably be defined in the class. You can remove the parameter and get xp from the item in the smelt method

src/targets/furnace.js
```diff
@@ -0,0 +1,98 @@
...
+                    "You need at least 1 copper and 1 tin in your inventory."
```
EthanVieira: Maybe mention that a bronze bar requires this. "You need at least one copper and one tin ore to smelt a bronze bar." Also not sure what we've been doing so far, but I generally like spelling out numbers if I know they are less than ten.

src/ui/panels/inventory.js
```diff
@@ -79,33 +79,28 @@ export class Inventory {
...
     getInventoryIndex(itemName) {
-        let playerItems = characterData.getInventory();
-        for (let i = 0; i < playerItems.length; i++) {
-            if (Object.keys(playerItems[i]).length && playerItems[i].item == itemName) {
-                return i;
-            }
-        }
-
-        return -1;
+        return characterData.getInventory().findIndex((item) => item.item == itemName);
```
EthanVieira, line 82: 🚀

src/ui/panels/inventory.js
```diff
@@ -79,33 +79,28 @@ export class Inventory {
...
 
     // Returns first instance of an item in inventory
     // that contains the given keyword in its name.
     // The first skill included in skillsRequired will
     // be used for sorting inventory items that match the keyword.
     async getKeywordInInventory(keyword, mustBeUsable = false, skillsRequired = []) {
-        let playerItems = characterData
-            .getInventory()
-            .filter((item) => Object.keys(item).length && item.item.includes(keyword));
+        if (mustBeUsable) {
```
EthanVieira, line 90: good call

src/ui/panels/inventory.js
```diff
@@ -129,11 +124,11 @@ export class Inventory {
...
-                    .getInventory()
-                    .indexOf(itemClasses[i].constructor.name);
             }
+        } else {
+            return characterData
+                .getInventory()
+                .findIndex((item) => item.item.includes(keyword));
         }
 
         return -1;
```
EthanVieira, line 134: this will never be reachable now. can remove

EthanVieira (changes requested):
Nice work - seems to work well. I am somewhat questioning your picture of the furnace including a person smelting, but I find it kind of hilarious so idk. 

Also can you pull in master to make sure nothing else broke.

src/targets/furnace.js
EthanVieira: Now that I think about it, the error message and the required ores should also be specified in the class. I think you should be able to reduce the smelt method to only one parameter and get everything else from the item class.

src/targets/furnace.js
CurtisGreen: Yeah nice catch

src/targets/furnace.js
CurtisGreen: 1) I tried that first but I leveled up so quickly. I guess that doesn't take into account time taken to get ore though...
2) That made sense for bones since you bury them directly but I'm not sure if it makes sense here, xp isn't stored in the ore or logs, it's stored in the rock/tree object which is most similar to Furnace here

src/ui/panels/inventory.js
CurtisGreen: Alrighty

src/README.md
CurtisGreen: I think whenever you need to modify multiple areas of code to do something it should be documented. I get tilted at work when I make a change but it turns out there's some other part that I need to update that's not obvious

src/ui/panels/inventory.js
CurtisGreen: Actually looking at it I think that -1 is needed. There are several conditions in the upper part of the function that if they are false it would go to that -1 at the end.

src/targets/furnace.js
CurtisGreen: Updated xp to 6. You haven't convinced me on putting the xp in the bronze bar class yet though

src/targets/furnace.js
CurtisGreen: Erorr messages updated here. https://github.com/EthanVieira/rs-clicker/pull/60/commits/f58a102e21838136430a3019f6ce252ac0ec0b07

src/targets/furnace.js
CurtisGreen: Done https://github.com/EthanVieira/rs-clicker/pull/60/commits/f58a102e21838136430a3019f6ce252ac0ec0b07

src/targets/furnace.js
CurtisGreen: Done https://github.com/EthanVieira/rs-clicker/pull/60/commits/f58a102e21838136430a3019f6ce252ac0ec0b07

src/targets/furnace.js
CurtisGreen: Done https://github.com/EthanVieira/rs-clicker/pull/60/commits/f58a102e21838136430a3019f6ce252ac0ec0b07

src/ui/panels/inventory.js
EthanVieira: Ah yes I see what you mean

src/targets/furnace.js
EthanVieira: I think the xp should be stored somewhere so you can reference it as the xp gained from smelting a bronze bar (and not as some random number) (same with error statement and required ores). It makes sense to store the xp in the single entity that most represents that specific action. The specific action of creating a bronze bar is represented more perfectly by the bronze bar itself than it is by the furnace because the furnace also represents the action of creating all other types of bars. Storing in the class would encapsulate information that one should not need to know when calling the smelt method. It reduces the complexity overall of the smelt method and this case statement. Calling smelt("BronzeBar") is so much cleaner.

src/README.md
EthanVieira: Ehh I won't make you remove it but I don't like this section. If you are adding stuff you should work through the logic and understand the full flows. Having a cheat sheet like this seems lazy and annoying to maintain. Depending on what needs to be added, it may not even be relevant or it may sway a developer to force his code into the prescribed buckets instead of taking a better solution.

src/targets/furnace.js
CurtisGreen: Done https://github.com/EthanVieira/rs-clicker/pull/60/commits/d0e349d408a86d7b07a378c05bf957c70fb030cc

src/README.md
CurtisGreen: I think it's pretty trivial to update and in cases where you need to update files that are not in the same area they can easily be overlooked

EthanVieira (approved):

PR #61, opened 2021-04-25
Standardization

CurtisGreen:
- show/hide => setVisible(bool = true)
- Empty inventory/equipment slots are now `null` instead of `{}`
- `let` => `const` when possible
- Cleaned up array functions in equipment/inventory
- All weapons now have their own parent classes (Ex: `BronzeDagger extends Dagger`
- Parent classes now reside in their respective folders rather than at the same level of them
    - Ex: dagger.js resides in the dagger folder rather than outside of it
- Parent classes now contain animation info (rather than everything being in `level.js`)


src/items/equipment/bows/bronze-crossbow.js
```diff
@@ -1,11 +1,6 @@
-import Equipment from "../equipment.js";
-import { EQUIPMENT } from "../../constants/constants.js";
-
-export default class BronzeCrossbow extends Equipment {
-    // Attack types
-    slot = EQUIPMENT.SLOTS.WEAPON;
-    skill = EQUIPMENT.WEAPON_TYPES.RANGED;
+import Bow from "./bow.js";
 
+export default class BronzeCrossbow extends Bow {
```
EthanVieira: Might want a crossbow class. They do use bolts instead of arrows so the animation would be different. Idk if we even care that much about crossbows though

src/items/equipment/knives/knife.js
```diff
@@ -0,0 +1,19 @@
...
+import { EQUIPMENT } from "../../../constants/constants.js";
+
+export default class Knife extends Equipment {
+    // Attack types
+    slot = EQUIPMENT.SLOTS.WEAPON;
+    skill = EQUIPMENT.WEAPON_TYPES.RANGED;
+
+    // Attack animation
+    animation = {
+        imageName: "bronze-arrow",
```
EthanVieira: Should definitely be using a knife here although may be out of scope for this

EthanVieira (changes requested):
Nice work it's lookin real spiffy. Can we do similar organization for trees under targets/resources?

src/items/equipment/knives/knife.js
CurtisGreen: Are there a lot of animations that need to be added/updated or just these two? If it's just these two I might do them, otherwise I'd do animations as another PR I think

EthanVieira: I _think_ these are the only ones that need updating now

src/items/equipment/bows/bronze-crossbow.js
CurtisGreen: Done https://github.com/EthanVieira/rs-clicker/pull/61/commits/f56516055f88cc3579f2713c3aac1ed45cda1cb1

EthanVieira (approved):

PR #62, opened 2021-04-25
Bug fixes

EthanVieira:
* Right click menu no longer stays open (with the exception of an unlikely case that I documented)
* Scroll window now works on quest list and clan

* Since the quest list scroll window is shaped differently from others, a full list will go off expected area a little bit. Will hopefully resolve later


src/ui/panels/clan.js
```diff
@@ -39,6 +39,8 @@ export class Clan {
             name: "clans",
             x: 535,
             y: 280,
+            textStartOffsetX: 30,
+            textStartOffsetY: 0,
```
CurtisGreen: This seems to be 30 & 0 most of the time. Maybe make those the default values and use it when the user doesn't pass them in so you don't have to pass these every time you make a scroll window w/ text in it.

Maybe something like:
```js
if (format.textStartOffsetX == undefined) {
   format.textStartOffsetX = 30;
}
```

CurtisGreen (changes requested):

src/ui/scroll-window.js
```diff
@@ -191,6 +196,12 @@ export class ScrollWindow extends Phaser.Scene {
                         this.scroll();
                     }
                 });
+
+                // If refreshing when scrolled down, update to previously scrolled position
+                let deltaY =
+                    (this.curScrollBarY - this.scrollBar.y) /
+                    (remainingScrollBarDist / remainingListDist);
+                this.scroll(deltaY);
```
CurtisGreen: This seems to mess up the chat window when examining things. What's the purpose of it?

EthanVieira: It preserves the spot of the scroll bar when it is refreshed. Needed for quest list because if you were trying to scroll down when enemies were being killed by autoclickers, it would refresh to the top at every kill and it would be impossible with fast kills. I'll throw in a flag to determine whether or not it should preserve it to fix this.

src/ui/panels/clan.js
EthanVieira: Updated! https://github.com/EthanVieira/rs-clicker/pull/62/commits/48f5a01bca18681a5c3316b66b087033ee369f1b

src/ui/scroll-window.js
EthanVieira: Updated! https://github.com/EthanVieira/rs-clicker/pull/62/commits/8074d0484ceb6b37e11d37098e47e3539963a00e

CurtisGreen (approved):

PR #63, opened 2021-04-30
add stats to quest window

EthanVieira:
image


src/ui/panels/quests.js
```diff
@@ -0,0 +1,232 @@
+import { CONSTANTS } from "../../constants/constants.js";
```
CurtisGreen, line 1: FYI, if you rename the file then do no changes (or only minor changes) then git add the "deleted" file and the new file it will register as a name change rather than deletion + creation.

src/ui/panels/quests.js
```diff
@@ -0,0 +1,232 @@
...
+                            0,
+                            0,
+                            printedAmount +
+                                "/" +
+                                questAmount +
+                                " " +
+                                prettyPrintCamelCase(enemy) +
+                                "s",
+                            {
+                                fill: enemiesKilled >= questAmount ? "#00ff00" : "yellow",
```
CurtisGreen: Try changing this formatting to RS font and see if it still looks good

CurtisGreen (changes requested):
image
Add this change to make things look nicer

src/ui/panels/quests.js
EthanVieira: Did not look good unfortunately

EthanVieira:
> Add this change to make things look nicer

https://github.com/EthanVieira/rs-clicker/pull/63/commits/abcdf58e0ca5212b17a2b2cf4b58d9a6bb4feb15 updated! Looks so much better

src/ui/panels/quests.js
EthanVieira: 🤔

CurtisGreen: As a separate commit I mean

src/ui/panels/quests.js
```diff
@@ -158,7 +158,7 @@ export class Quests {
             let sceneText = this.statsScrollWindow.add
                 .text(0, 0, statTexts[statkey], {
                     fill: "white",
-                    fontSize: 11,
+                    font: "14px runescape",
```
CurtisGreen: I think bumping it up to 16 or 16.5 would be good. Also btw if it's not too wide I'm pretty sure you can put "white" into font rather than having it separate.

CurtisGreen: In addition maybe separate text and number text so that you can left align all the numbers. Then the numbers can be truncated by # of digits to fit within the space using k, m, b, etc. For easy reading maybe color the numbers orange or blue like the right click menu or chat menu.

CurtisGreen (changes requested):

src/ui/panels/quests.js
EthanVieira: I think 16 is too big. Left at 14. Updated quest and stat text here: https://github.com/EthanVieira/rs-clicker/pull/63/commits/57c5af2927df68135c12811f90015b8c5e87b72c

src/ui/panels/quests.js
```diff
@@ -0,0 +1,243 @@
...
+    refreshStats() {
+        this.statsScrollWindow.clearObjects();
+        let statTexts = this.dashboard.scene.get(CONSTANTS.SCENES.STATS).getStats();
+
+        for (const statkey in statTexts) {
+            let row = new TextRow(this.statsScrollWindow, 0, 0, []);
+
+            let statText = this.statsScrollWindow.add
+                .text(0, 0, statTexts[statkey]["text"], {
+                    font: "14px runescape white",
```
CurtisGreen: `font: "16px runescape"`

Looks like I was wrong about being able to include fill color in font. This fails so it renders as normal font. And since it's RS font bump it up to 16px to match the other one

src/ui/panels/quests.js
```diff
@@ -0,0 +1,243 @@
...
+                })
+                .setDepth(3);
+            row.add(statText);
+
+            let statTxtAmount = this.statsScrollWindow.add
+                .text(100, 0, getItemText(statTexts[statkey]["amount"])[0], {
+                    font: "14px runescape white",
```
CurtisGreen:
```js
font: "16px runescape",
fill: "orange",
```

Same thing here, easier to read with the numbers orange imo, matches the right click menu

CurtisGreen (changes requested):
Also dps in the stats window doesn't update all the time to match dps in the main window, it only updates on click

src/ui/panels/quests.js
EthanVieira: updated: https://github.com/EthanVieira/rs-clicker/pull/63/commits/294623b39a2c655d7ba30fc8181b48fd4e8f8a2c

src/ui/panels/quests.js
EthanVieira: updated: https://github.com/EthanVieira/rs-clicker/pull/63/commits/294623b39a2c655d7ba30fc8181b48fd4e8f8a2c

CurtisGreen (approved):

PR #64, opened 2021-05-01
Hammer anvil

AndrewVieira:
New Features and Changes

- Added Varrock Anvil Location
- Can smith bronze daggers at anvil location, requires hammer and bronze
- Added smithable class for equipment, use it for equipment that can be smithed
- Add hasItem() in utlities, checks if item with given name is in inventory (used to check for hammer)
- Changed some 'var's to 'let's and 'const's


src/items/equipment/smithable.js
```diff
@@ -0,0 +1,7 @@
+import Equipment from "./equipment.js";
```
EthanVieira, line 1: Haven't told you about this but we are using prettier to auto-format. Not necessarily a requirement to get it set up, but might be something to look into. You can see that other imports are always using brackets.

src/targets/anvil.js
```diff
@@ -0,0 +1,104 @@
...
+        const cameraHeight = scene.cameras.main.height;
+        const width = 220;
+        const height = 140;
+        const x = cameraWidth / 2 - width + 60;
+        const y = cameraHeight / 2 - height + 40;
+
+        // Add invisible button for anvil
+        this.sprite = new Button(scene, x, y, width, height);
+        this.sprite.on("pointerdown", (pointer) => {
+            if (pointer.rightButtonReleased() && !pointer.leftButtonDown()) {
```
EthanVieira: Right click menu doesn't seem to work, and just realized it doesn't work on the furnace level. Not sure what's wrong there @CurtisGreen

src/utilities.js
```diff
@@ -125,6 +125,22 @@ export function runOnLoad(scene, func) {
     }
 }
 
+// Returns true if item exist in inventory
+export function hasItem(inventory, itemName) {
```
EthanVieira: We have some methods in inventory.js that you should be able to use for this.

src/targets/anvil.js
```diff
@@ -0,0 +1,104 @@
...
+        if (hasHammer === false) {
+            chat.writeText("You need a hammer to work the metal with.");
+            return;
+        }
+
+        const selectedItem = inv.inventory[selectedIndex];
+
+        switch (selectedItem.name) {
+            case "Bronze Bar":
+                this.smith("BronzeDagger");
```
EthanVieira, line 60: Just putting a note here for our future selves when we have more smithing recipes - Setting the smithing recipe should be done by right clicking the anvil so that it doesn't have to be chosen on every click. This will probably change to this.smith("BronzeItem") and the smith method will check the anvil state.

src/targets/anvil.js
```diff
@@ -0,0 +1,104 @@
...
+                break;
+            default:
+                chat.writeText("The anvil can only be used with the bar selected");
+                break;
+        }
+    }
+
+    // Take ore and turn it into a bar
```
EthanVieira: Bar(s) into equipment*

src/targets/anvil.js
```diff
@@ -0,0 +1,104 @@
...
+    async smith(weaponName) {
```
EthanVieira: Change the arg name here to be itemName or equipmentName as we can smith more than just weapons. When we have more recipes it'll probably be more like barType but that doesn't make sense right now

src/targets/anvil.js
```diff
@@ -0,0 +1,104 @@
...
+        const inv = this.scene.dashboard.inventory;
+        const chat = this.scene.scene.get(CONSTANTS.SCENES.CHAT);
+        const bronzeDagger = await getItemClass(weaponName, this.scene.dashboard);
```
EthanVieira: Go ahead and change this variable name to something more equipment-generic so we don't have to in the future

src/targets/anvil.js
```diff
@@ -0,0 +1,104 @@
...
+
+        const indices = bronzeDagger.bars.map((bar) => inv.getInventoryIndex(bar));
+        const allExist = indices.every((index) => index >= 0);
+
+        // All ingredients are in inventory
+        if (allExist) {
+            const added = inv.addToInventory(bronzeDagger, true);
+
+            // Inventory has space, reduce ore counts by 1
```
EthanVieira: reduce bar counts by the required amount*

src/targets/anvil.js
```diff
@@ -0,0 +1,104 @@
...
+            if (added) {
+                indices.forEach((index) => {
+                    const obj = inv.inventory[index];
+                    obj.setNumItems(obj.numItems - 1);
```
EthanVieira: the 1 here should be pulled from the item's required amount

src/targets/anvil.js
```diff
@@ -0,0 +1,104 @@
...
+                });
+
+                // Log click for stats
+                this.scene.stats.updateClickedTargetStat();
+                characterData.addSkillXp("smithing", bronzeDagger.xp);
+                this.scene.enemyKilled("bronzeDagger");
```
EthanVieira: Should generalize the xp and item name here

src/targets/anvil.js
```diff
@@ -0,0 +1,104 @@
...
+
+                // Show smith animation
+                this.scene.clickAnimation();
+            }
+        } else {
+            chat.writeText(bronzeDagger.smeltingErrorMessage);
```
EthanVieira: generalize here

EthanVieira (changes requested):
Nice job! Overall looks p good. Left some small comments. We _might_ want to move the bg picture up a bit to be forward thinking as we transition all levels to not have the main targetable be blocked by the chat box.

Also looks like the Bronze Dagger text in the quest list is too long, but we still need the space on the right for potential 4 digit quests, so we might wanna look into resizing the font there

src/targets/anvil.js
CurtisGreen: Just checked it out, is there a reason it's `rightButtonReleased()`? It works as expected with `rightButtonDown`

EthanVieira: I may have accidently introduced that when fixing other right click bugs. It should be rightButtonDown

CurtisGreen: Alrighty, Andy when you swap this one can you swap it on furnace.js as well?

AndrewVieira: Yes I will

src/items/tools/hammer.js
```diff
@@ -0,0 +1,27 @@
+import Tool from "./tool.js";
+
+export default class Hammer extends Tool {
```
CurtisGreen, line 3: When you use a hammer on a bar it should tell you that you need to go to an anvil to do this. Right now it just says the default "this is an invalid crafting combination"

src/scenes/world-map.js
```diff
@@ -72,6 +72,15 @@ export class WorldMapScene extends Phaser.Scene {
                 y: MAP.AL_KHARID_FURNACE.Y,
                 key: CONSTANTS.SCENES.AL_KHARID_FURNACE,
             },
+            {
+                text: "Varrock Anvil",
+                condition: characterData.getQuestCompleted(
+                    CONSTANTS.PREREQUISITES.VARROCK_ANVIL
```
CurtisGreen, line 78: For whatever reason this never resolves to true and I can't get in the anvil level. It's possible al kharid is never set to true for some reason

CurtisGreen (changes requested):

src/scenes/world-map.js
AndrewVieira: Did you mean you can't go into the "Varrock Anvil" level after making 15 bronze bars? I'm not having any issues on my end.

CurtisGreen: Oh I didn't realize there was a number of bars minimum

src/items/equipment/smithable.js
AndrewVieira: I used prettier, but it wanted to change almost every file when I ran it (115 files.) I only applied the changes where text actually changed.

src/targets/anvil.js
AndrewVieira: I added a card to Trello under planned features as a reminder, since this is not within my planned scope for this pull.

src/items/tools/hammer.js
AndrewVieira: Has been implemented with recent commits.

src/items/equipment/smithable.js
CurtisGreen: We have a prettier config in the repo, make sure your prettier is picking it up

.prettierrc.js
```diff
@@ -1,4 +1,4 @@
 module.exports = {
     tabWidth: 4,
-    printWidth: 90
+    printWidth: 90,
```
CurtisGreen, line 3: Prettier running on itself lol

src/items/tools/hammer.js
```diff
@@ -0,0 +1,40 @@
...
+
+    getRecipe(itemName) {
+        let output = {
+            className: "",
+            numRequiredItems: 0,
+            xpGiven: 0,
+            requires: "",
+        };
+
+        switch (itemName) {
```
CurtisGreen, line 26: These can be stacked:
```js
case "Bronze Bar":
case "Iron Bar":
case "Steel Bar":
    output.requires = "Anvil";
    break;
```

src/items/equipment/smithable.js
AndrewVieira: I believe it is. The reason prettier is changing so many files is that it wants to change the line endings from "LF" to "CRLF" on almost every file.

src/scenes/varrock-anvil.js
```diff
@@ -0,0 +1,25 @@
+import { CONSTANTS } from "../constants/constants.js";
+import { LevelScene } from "./level.js";
+import { Anvil } from "../targets/anvil.js";
+
+export class VarrockAnvilScene extends LevelScene {
+    constructor() {
+        super({
+            levelType: CONSTANTS.LEVEL_TYPE.SMITHING,
```
CurtisGreen, line 8: Need to add this to the check in `AutoClicker::start()`.  to prevent getting an error when autoclickers/clan members try to "click" on the target.
image

image

CurtisGreen (changes requested):

.prettierrc.js
AndrewVieira: Am I running the prettier wrong? I use the command "npx prettier --write . "

src/targets/anvil.js
```diff
@@ -0,0 +1,110 @@
...
+                // Log click for stats
+                this.scene.stats.updateClickedTargetStat();
+                characterData.addSkillXp("smithing", item.xp);
+                this.scene.enemyKilled(item.questName);
+
+                // Show smith animation
+                this.scene.clickAnimation();
+            }
+        } else {
+            chat.writeText(item.smeltingErrorMessage);
```
EthanVieira: I think this should be smithingErrorMessage

EthanVieira (changes requested):

src/targets/anvil.js
AndrewVieira: I made the change

src/scenes/varrock-anvil.js
AndrewVieira: I was able to fix this I think? I was able to stop the error with the bot. I'm not sure 100% how auto-clickers work yet

src/items/tools/hammer.js
AndrewVieira: I made the change

EthanVieira (approved):

CurtisGreen (approved):

.prettierrc.js
CurtisGreen: Are you running it while in the same directory as this file? Regardless, PR looks good now so it's approved

PR #65, opened 2021-05-08
XP animation

CurtisGreen:
## Features
- Now shows animation for xp per skill
- Shows skill icon + grey circle background

## Code organization
- Created an animation scene that runs in the background like the audio scene and will take care of animations going forward
- Moved click target animation logic into that class
- Made the target responsible for the animation config data


src/ui/animation.js
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        const startX = 470;
+        const startY = Math.floor(this.cameras.main.height / 2);
+        const endX = startX;
+        const endY = 20;
+
+        // Gold runescape font
+        const font = {
+            font: "24px runescape",
```
EthanVieira: I think it looks better with font size of 18px and the startY value at 100

EthanVieira (changes requested):
There is a bug where if you sell your tool on a resource level, it will still keep using the tool until some other state is reached. I haven't tested it too much, but it has something to do with having an item equipped i think?

Won't make you do it here, but eventually I think we will want to have a total xp counter right above the animation that is toggled by the xp icon near the map.

One thing that I'm not sure whether or not was happening before is that it got pretty laggy when I had a good amount of autoclickers. We should probably limit autoclickers to one of each type and then have upgrades for them.

CurtisGreen:
> One thing that I'm not sure whether or not was happening before is that it got pretty laggy when I had a good amount of autoclickers. We should probably limit autoclickers to one of each type and then have upgrades for them.

I could be wrong but I think the pixelArt setting that made everything look nicer took a toll on performance. Having upgradeable auto clickers is something I've wanted to do in the future so I think that's a good way to prevent this issue

src/ui/animation.js
CurtisGreen: Done here: https://github.com/EthanVieira/rs-clicker/pull/65/commits/f1e491a0b907e0df873ae34268d4d6cac09031f5

CurtisGreen:
> There is a bug where if you sell your tool on a resource level, it will still keep using the tool until some other state is reached. I haven't tested it too much, but it has something to do with having an item equipped i think?

Fixed here: https://github.com/EthanVieira/rs-clicker/pull/65/commits/1368cd0200855662b292edd3e3035083017121ca

EthanVieira (approved):

PR #67, opened 2022-04-05
Quest points

EthanVieira:
* Completing a level now gives quest points which are visible in the quest tab

### DIFF
--- a/src/cookie-io.js
+++ b/src/cookie-io.js
@@ -170,7 +170,6 @@ class CharacterData {
     setQuestCompleted(scene) {
         if (this.checkScene(scene)) {
             this.characterData.levels[scene].questCompleted = true;
-
             // Unlock levels
             for (let level in this.characterData.levels) {
                 if (
@@ -188,6 +187,14 @@ class CharacterData {
                 }
             }
         }
+    }
+
+    addQuestPoints(points) {
+        this.characterData.questPoints += points;
+    }
+
+    getQuestPoints() {
+        return this.characterData.questPoints;
     }
 
     calcQuestTier(numKilled, tiers) {
--- a/src/cookie-io.js
+++ b/src/cookie-io.js
@@ -170,6 +170,7 @@ class CharacterData {
     setQuestCompleted(scene) {
         if (this.checkScene(scene)) {
             this.characterData.levels[scene].questCompleted = true;
+
             // Unlock levels
             for (let level in this.characterData.levels) {
                 if (
--- a/src/ui/panels/quests.js
+++ b/src/ui/panels/quests.js
@@ -13,6 +13,7 @@ export class Quests {
     statsTabButton;
     questsScrollWindow;
     statsScrollWindow;
+    questPointsText;
 
     currentTab;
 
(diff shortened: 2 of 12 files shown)


src/cookie-io.js
```diff
@@ -170,7 +170,6 @@ class CharacterData {
     setQuestCompleted(scene) {
         if (this.checkScene(scene)) {
             this.characterData.levels[scene].questCompleted = true;
-
```
CurtisGreen: I personally like leaving an empty line above comments so they function as headers, not the end of the world though

src/ui/panels/quests.js
```diff
@@ -188,27 +196,29 @@ export class Quests {
             for (var enemy in enemies) {
                 let enemiesKilled = characterData.getEnemiesKilled(level, enemy);
 
-                for (
-                    var tier = 1;
-                    tier <=
-                    characterData.calcQuestTier(enemiesKilled, scene.questAmounts[enemy]);
-                    tier++
-                ) {
+                let questTier = characterData.calcQuestTier(
```
CurtisGreen: While we're adding new code we should try to be more functional and only use `let` when we're actually changing the value and `const` otherwise

src/ui/panels/quests.js
```diff
@@ -62,6 +63,13 @@ export class Quests {
                 }
             });
 
+        this.questPointsText = dashboard.add
+            .text(645, 233, String(characterData.getQuestPoints()), {
```
CurtisGreen, line 67: I think it auto casts to String but you can double check

CurtisGreen (changes requested):

src/ui/panels/quests.js
```diff
@@ -188,27 +196,29 @@ export class Quests {
...
-                    tier <=
-                    characterData.calcQuestTier(enemiesKilled, scene.questAmounts[enemy]);
-                    tier++
-                ) {
+                let questTier = characterData.calcQuestTier(
+                    enemiesKilled,
+                    scene.questAmounts[enemy]
+                );
+
+                for (var tier = 1; tier <= questTier; tier++) {
```
CurtisGreen: I know it was already there but `var` -> `let`

src/cookie-io.js
EthanVieira: Reverted! https://github.com/EthanVieira/rs-clicker/pull/67/commits/61dfc063b2991372be5e92829d7e8ae2813cf74a

src/ui/panels/quests.js
EthanVieira: Changed some of them https://github.com/EthanVieira/rs-clicker/pull/67/commits/61dfc063b2991372be5e92829d7e8ae2813cf74a

src/ui/panels/quests.js
EthanVieira: Removed the cast https://github.com/EthanVieira/rs-clicker/pull/67/commits/61dfc063b2991372be5e92829d7e8ae2813cf74a

src/ui/panels/quests.js
EthanVieira: https://github.com/EthanVieira/rs-clicker/pull/67/commits/61dfc063b2991372be5e92829d7e8ae2813cf74a

CurtisGreen (approved):

PR #69, opened 2022-04-07
Create attack style panel

CurtisGreen:
# Attack Style Panel
image

Simple PR to add attack style panel, make the buttons interactable and store their values in the cookie. Doesn't currently change XP gains between attack/strength/defense as we still need to decide how we'll handle those in the future. Currently plan is to make all of the buttons on the dashboard (inventory, skills, etc.) clickable


src/scenes/load.js
```diff
@@ -137,8 +137,29 @@ export class LoadScene extends Phaser.Scene {
...
         this.load.image("scroll-footer", "src/assets/ui/ScrollFooter.png");
         this.load.image("scroll-button", "src/assets/ui/ScrollButton.png");
+        this.load.image(
+            "attack-style-button",
+            "src/assets/ui/buttons/AttackStyleButton.png"
+        );
+        this.load.image("attack-style-panel", "src/assets/ui/AttackStylePanel.png");
+        this.load.image(
+            "attack-style-1-button",
+            "src/assets/ui/buttons//AttackStyle1Button.png"
```
AndrewVieira: There's two forward slashes here

AndrewVieira (commented):
The code seems good to me. I tested it out and I had no issues. I like that the panel keeps the attack style the same when changing settings or refreshing the page.

Good work so far!

src/scenes/load.js
CurtisGreen: Fixed https://github.com/EthanVieira/rs-clicker/pull/69/commits/602083f85d2aea2b594e1370226da5e3845838b1

EthanVieira (approved):
Nice job!

PR #70, opened 2022-04-30
Bronze age

AndrewVieira:
Hi there are a few things that need work still, but I wanted someone to code review my work so far.

WHAT'S NEW
1. When smithing, you have three choices that pop up in a modal window. The first green box is for bronze dagger, 2nd is the bronze sword, third is the bronze scimitar. Press the small white box to close without choosing. This shapes are just stand ins for now. I will use real images later.
2. Right click anvil to change your choice with "Choose" option
3. A new modal class for choosing recipes (like smithing)

WHAT NEEDS REVIEW
1. Any code that can be improved
2. Any bugs you find
4. What code should stay in modal-window,.js and what code should go to future child class smithing-interface.js
5. How do I use images for my modal class? (I tried using sprites but nothing showed up)
6. Anything else needed

KNOWN BUGS
Hammer hand animation works when smithing dagger, but not sword or scimitar


src/ui/modal-window.js
```diff
@@ -0,0 +1,96 @@
...
+            this.elements[i].destroy();
+        }
+        
+        this.elements = [];
+        this.elementEvents = [];
+    }
+
+    addElements(elements) {
+        let offset = 8;
+        this.elementEvents = elements.map((obj) => obj);
```
CurtisGreen: What's the purpose of this, to copy them? If so `this.elementEvents = [...elements];` is functionally equivalent

src/ui/modals/smithing-interface.js
```diff
@@ -0,0 +1,97 @@
+/* import { ModalWindow } from "../modal-window.js";
```
CurtisGreen: I think we should delete the file instead of just commenting stuff out, can look back at git history if we need the code again

src/ui/modal-window.js
```diff
@@ -0,0 +1,96 @@
...
+
+        for (let i = 0; i < elements.length; i++) {
+            
+            let elementButton = new Button(this.scene, this.x + offset, this.y + 16, 48, 48);
+            elementButton.on("pointerup", () => {
+                if (this.visible == true) {
```
CurtisGreen: Nitpick
```suggestion
                if (this.visible) {
```

src/ui/modal-window.js
```diff
@@ -0,0 +1,96 @@
...
+        
+        this.elements = [];
+        this.elementEvents = [];
+    }
+
+    addElements(elements) {
+        let offset = 8;
+        this.elementEvents = elements.map((obj) => obj);
+
+        for (let i = 0; i < elements.length; i++) {
```
CurtisGreen: If you don't need the index you can do this:
```suggestion
        for (let element of elements) {
```

CurtisGreen: Same for other indexed loops

CurtisGreen (changes requested):
Code pretty good so far just a few minor style changes. I haven't tested it yet, will try to do that tomorrow 👍

src/targets/anvil.js
```diff
@@ -27,46 +31,85 @@ export class Anvil extends ClickableObject {
         // Add invisible button for anvil
         this.sprite = new Button(scene, x, y, width, height);
         this.sprite.on("pointerdown", (pointer) => {
-            if (pointer.rightButtonDown() && !pointer.leftButtonDown()) {
+            if (pointer.rightButtonDown() && !pointer.leftButtonDown() && this.modalWindow.visible == false) {
```
CurtisGreen:
```suggestion
            if (pointer.rightButtonDown() && !pointer.leftButtonDown() && !this.modalWindow.visible) {
```

src/ui/modal-window.js
```diff
@@ -0,0 +1,96 @@
...
+                }
+            });
+
+            let elementIcon = this.scene.add.graphics();
+            elementIcon.setDepth(3);
+            elementIcon.fillStyle(0x22EE22);
+            elementIcon.fillRect(this.x + offset, this.y + 16, 48, 48);
+            elementIcon.visible = this.visible;
+
+            offset += 56;
```
CurtisGreen: Would be better if offset was a constant based off of the loop index rather than adding to it each iteration. Makes it more obvious what its value is

AndrewVieira:
Thanks for reviewing the code. I will start making changes tomorrow.

src/items/equipment/scimitars/bronze-scimitar.js
```diff
@@ -1,6 +1,11 @@
 import Scimitar from "./scimitar.js";
 
 export default class BronzeScimitar extends Scimitar {
+    xp = 30.0;
```
EthanVieira: Gives 25xp in RS and requires 2 bars

src/items/equipment/swords/bronze-sword.js
```diff
@@ -1,6 +1,11 @@
 import Sword from "./sword.js";
 
 export default class BronzeSword extends Sword {
+    xp = 30.0;
```
EthanVieira: 12.5xp

src/targets/anvil.js
```diff
@@ -27,46 +31,85 @@ export class Anvil extends ClickableObject {
...
                 this.clickTarget();
             }
         });
+
+        //Create modal window
+        this.modalWindow = new ModalWindow(this.scene);
     }
 
     async clickTarget() {
+        if (this.modalWindow.visible == true) {
```
EthanVieira: if (this.modalWindow.visible)

src/targets/anvil.js
```diff
@@ -27,46 +31,85 @@ export class Anvil extends ClickableObject {
...
         if (hasHammer < 0) {
             chat.writeText("You need a hammer to work the metal with.");
+            return false;
+        }
+
+        return true;
+    }
+
+    async selectRecipe(itemName) {
+        if (this.hasMaterials() == false) {
```
EthanVieira: if (!this.hasMaterials())

src/targets/anvil.js
```diff
@@ -27,46 +31,85 @@ export class Anvil extends ClickableObject {
...
+                return;
         }
+
+        this.modalWindow.addElements(elements);
+        this.modalWindow.setVisible(true);
     }
 
     // Take bar and turn it into a smithable item
     async smith(itemName) {
+        if (this.hasMaterials() == false) {
```
EthanVieira: if (!this.hasMaterials())

src/targets/anvil.js
```diff
@@ -27,46 +31,85 @@ export class Anvil extends ClickableObject {
...
+
         const selectedItem = inv.inventory[selectedIndex];
+        
+        let elements = [];
+        this.modalWindow.clearElements();
 
         switch (selectedItem.name) {
             case "Bronze Bar":
-                this.smith("BronzeDagger");
+                    elements = ["BronzeDagger", "BronzeSword", "BronzeScimitar"];
```
EthanVieira: I think I'm still in favor of having all of the modal window specifics inside of a child modal window class. i.e. bronze-smithing-modal-window.js or if all of the different smithing modals will be similar then maybe a smithing-modal-window that takes "bronze" as an argument. 

I prefer it that way for two reasons:
1) the modal window will be used for other things like the shop which, in addition to having different elements, could have a different background, multiple tabs, whether or not it has a scrollbar etc. Having to insert all that configurable info every time we create a modal window would be messy as opposed to just a single statement creating a child modal window class that has all the info inside.
2) the OOP model is so deeply ingrained in everything else we are doing it might as well continue

@CurtisGreen thoughts?

EthanVieira (changes requested):
Naming convention for Bronze spear and warhammer icons should follow the rest. 

Don't need "Detail" on BronzeFullHelmDetail.png".

BronzeArrowtips -> capitalize Tips (in models)

Something I want to start doing is having Data files that can be loaded in place of default-data.js so we can jump to certain parts of the game to easily test them. In the next pr, could you create a test-smithing-data.js (or something similar)? Basically just copy default-data.js and then add in the additional things that would be useful for testing smithing. It would have all the smithing locations unlocked and plenty of bars/ore in the inventory to start with.

For your question on how to use images with it - I would look at dashboard.js for that. There are a lot of images added there and some are set interactive to be buttons. You should be able to follow the same patterns.

Overall great job! I look forward to using this!

src/targets/anvil.js
AndrewVieira: I think your approach is best. I was overengineering a bit. Some future modal windows like shop and bank need to be a little customizable since the options can vary.

src/items/equipment/scimitars/bronze-scimitar.js
AndrewVieira: I think I got my data from an incorrect source. Could you tell me what source you are using?

EthanVieira: Use the OSRS wiki instead of the main RS wiki

src/targets/anvil.js
CurtisGreen: Hard to say until we make a new modal and see what we need/don't need to be generic

CurtisGreen (changes requested):
Finally got around to testing, sorry for being slow.

I see an error in the console logs here whenever I click on the anvil without having anything selected first.
https://i.imgur.com/6ITAMnJ.png 

Same error for the furnace
https://i.imgur.com/DdvQ89L.png

Error when clicking hammer then clicking anvil
https://i.imgur.com/EN096ab.png

Modal images didn't load properly. Maybe you didn't push them up?
https://i.imgur.com/t0EfRqv.png

AndrewVieira:
I fixed all the issues you guys mentioned and I fixed the animation bug too. I will be working on the modal class and adding images.

CurtisGreen:
> I fixed all the issues you guys mentioned and I fixed the animation bug too. I will be working on the modal class and adding images.

Sweet, when you get the chance can you resolve the comments you've fixed so we can know what has/hasn't been done

AndrewVieira:
PROGRESS PICTURE: https://imgur.com/a/gnmf1W5
I'm getting pretty close to being done.

WHAT'S NEW
1. Added images and cleaned up the interface (icon position can be moved around arbitrarily in code) 
2. Item is not smithable if not high enough smithing level or enough bronze bars. Item text changes black if not high enough level. Bar count text turns orange if not enough bars.
3. Adding smithing test data file to speed up testing

TO DO
1. Add remaining bronze smithing items and organize them.
2. Add comments in log if not enough bars or high enough level? (Do you want me to add this?)
3. Any other issues you guys find

Please let me know what needs fixing.

src/cookie-io.js
```diff
@@ -1,10 +1,12 @@
-import { defaultData } from "./default-data.js";
+// import { defaultData } from "./default-data.js";
+import { testSmithingData } from "./test-smithing-data.js"
```
CurtisGreen: You left some testing data in the commit

src/cookie-io.js
```diff
@@ -1,10 +1,12 @@
...
 import { CONSTANTS, FONTS } from "./constants/constants.js";
 import * as Utilities from "./utilities.js";
 
 function getDefaultData() {
     // Reset data (deep copy)
-    return JSON.parse(JSON.stringify(defaultData));
+    //return JSON.parse(JSON.stringify(defaultData));
+    return JSON.parse(JSON.stringify(testSmithingData));
```
CurtisGreen: here too

src/scenes/level.js
```diff
@@ -155,6 +155,11 @@ export class LevelScene extends Phaser.Scene {
     }
 
     enemyKilled(name) {
+        // Check if quest even exists
+        if (this.questAmounts[name] === undefined) {
```
CurtisGreen: 1) This can be simplified to `if(!this.questAmounts[name])` since undefined, null, and false all resolve to false in an if statement
2) If something is breaking by calling this function we should fix it rather than working around the bug. I'm fine with keeping it but add a console log under it describing the error like cookie-io does

src/targets/anvil.js
```diff
@@ -27,46 +33,79 @@ export class Anvil extends ClickableObject {
...
+    }
+
+    async hasMaterials() {
         const inv = this.scene.dashboard.inventory;
         const chat = this.scene.scene.get(CONSTANTS.SCENES.CHAT);
 
         // See if bar is selected
         const selectedIndex = inv.curSelectedItemIndex;
-        if (selectedIndex < 0) {
+        const selectedItem = inv.inventory[selectedIndex];
```
CurtisGreen, line 74: Moving the conditional down means we might do `inv.inventory[-1]`, need to check for that before getting `selectedItem`

CurtisGreen: Unless this no longer happens

src/targets/anvil.js
```diff
@@ -27,46 +33,79 @@ export class Anvil extends ClickableObject {
...
 
-        switch (selectedItem.name) {
-            case "Bronze Bar":
-                this.smith("BronzeDagger");
-                break;
-            default:
-                chat.writeText("The anvil can only be used with the bar selected");
-                break;
+    async selectRecipe(itemName) {
+        if (! await this.hasMaterials()) {
```
CurtisGreen: Just noticed this, why are the methods in this class and some of the other targets async? I don't see why it would be necessary since they don't await anything but maybe I'm misremembering something...

src/ui/modals/smithing-modal-window.js
```diff
@@ -0,0 +1,156 @@
...
+            element.destroy();
+        }
+        
+        this.elements = [];
+    }
+
+    getSpriteName(name) {
+        let result = name.toLowerCase();
+        result.toLowerCase();
+        result = result.replaceAll(" ", "-");
```
CurtisGreen: 1) You have an extra `result.toLowerCase()` there.
2) I think it's cleaner to put: `return name.toLowerCase().replaceAll(" ", "-");`
3) Is this useful for future modals? Might want to put it in the parent class

src/ui/modals/smithing-modal-window.js
```diff
@@ -0,0 +1,156 @@
...
+                    if (this.visible) {
+                        this.choice = element;
+                        this.setVisible(false);
+                    }
+                });
+
+                this.elements.push(elementButton);
+            }
+
+            let elementIcon = this.scene.add
```
CurtisGreen: const

src/ui/modals/smithing-modal-window.js
```diff
@@ -0,0 +1,156 @@
...
+                .image(positionX + 12, positionY, spriteName)
+                .setOrigin(0, 0)
+                .setDepth(3);
+            elementIcon.visible = this.visible
+
+            const textFont = levelFlag ? FONTS.SMITH_UNLOCKED : FONTS.SMITH_LOCKED;
+
+            let elementText = this.scene.add.text(positionX, positionY + this.iconHeight, item.item, textFont).setDepth(3);
```
CurtisGreen: const

src/ui/modals/smithing-modal-window.js
```diff
@@ -0,0 +1,156 @@
...
+            if (rowIndex >= this.maxIconsPerRow) {
+                positionX = this.x + this.leftOffset + this.horizontalPadding;
+                positionY += this.iconHeight + this.verticalPadding + 24;
+                rowIndex = 0;
+            }
+
+            const item = await getItemClass(element, this.scene.window);
+            const spriteName = await this.getSpriteName(item.name);
+
+            const barFlag = barSupply >= item.bars[0].count ? true : false;
```
CurtisGreen: Can be simplified to `const barFlag = barSupply >= item.bars[0].count;`

Same for one below this

src/ui/modals/smithing-modal-window.js
```diff
@@ -0,0 +1,156 @@
...
+                return;
+        }
+
+        let rowIndex = 0;
+
+        let positionX = this.x + this.leftOffset + this.horizontalPadding;
+        let positionY = this.y + this.topOffset + this.verticalPadding;
+
+        for (let element of elements) {
+            if (element === "SKIP") {
```
CurtisGreen: Maybe use `null` instead so you can do `if (!element)`

src/ui/modals/smithing-modal-window.js
```diff
@@ -0,0 +1,156 @@
...
+            }
+
+            const item = await getItemClass(element, this.scene.window);
+            const spriteName = await this.getSpriteName(item.name);
+
+            const barFlag = barSupply >= item.bars[0].count ? true : false;
+            const levelFlag = smithingLevel >= item.smithingLevel ? true : false;
+
+            if (barFlag && levelFlag) {
+                let elementButton = new Button(this.scene, positionX, positionY, this.buttonWidth, this.buttonHeight);
```
CurtisGreen: All of these objects can be changed to `const` from `let`

src/ui/modals/smithing-modal-window.js
CurtisGreen: I wonder if you can chain a `setVisible(this.visible)` on the end so it can be all in one

CurtisGreen (changes requested):

src/ui/modals/smithing-modal-window.js
```diff
@@ -0,0 +1,156 @@
...
+                .setOrigin(0, 0)
+                .setDepth(3);
+            elementIcon.visible = this.visible
+
+            const textFont = levelFlag ? FONTS.SMITH_UNLOCKED : FONTS.SMITH_LOCKED;
+
+            let elementText = this.scene.add.text(positionX, positionY + this.iconHeight, item.item, textFont).setDepth(3);
+            elementText.visible = this.visible;
+
+            let postfix = " bar";
```
CurtisGreen: `const postfix = (item.bars[0].count > 1) ? " bars" : " bar";`

src/ui/modals/smithing-modal-window.js
```diff
@@ -0,0 +1,156 @@
...
+        for (let element of elements) {
+            if (element === "SKIP") {
+                positionX += this.iconWidth + this.horizontalPadding;
+                rowIndex += 1;
+                continue;
+            }
+
+            if (rowIndex >= this.maxIconsPerRow) {
+                positionX = this.x + this.leftOffset + this.horizontalPadding;
+                positionY += this.iconHeight + this.verticalPadding + 24;
```
CurtisGreen: In general for loop creation logic like this I like to create these variables based off the index rather than adding to them.
Like instead of  `+=`  do:

```js
const yOffset = this.iconHeight + this.verticalPadding + 24;
const xOffset = ...
for (let [index, val] of array.entries()) {
    const positionY = index * yOffset;
    const positionX = index * xOffset;
    // ...
}
```

@EthanVieira Thoughts?

Another potential way of making this looping creation easier to read is mapping over the data and returning that into `elements` instead of creating elements and appending to it

src/ui/modals/smithing-modal-window.js
```diff
@@ -0,0 +1,156 @@
...
+        let result = name.toLowerCase();
+        result.toLowerCase();
+        result = result.replaceAll(" ", "-");
+        return result;
+    }
+
+    async setChoices(bar, barSupply, smithingLevel) {
+        this.clearChoices();
+
+        let elements = [];
```
CurtisGreen: Not sure of a better name but having `elements` and `this.elements` is pretty confusing

src/targets/anvil.js
AndrewVieira: I remember I was having issues without await here in this instance, but I don't remember the exact issue. I'll look at it again.

CurtisGreen: Yeah, maybe because this.hasMaterial is a sync as well? If you don’t want to deal with it in this PR I can look into it later

src/ui/modals/smithing-modal-window.js
AndrewVieira: I put it in the parent modal class for now. I was thinking we might just make it a general function and place it in utilities.js.

src/targets/anvil.js
AndrewVieira: I fixed the issue, but I found another potential issue for future reference. If you are making an item and run out of material, the selected index from inventory is still valid, so the selected item returned by inventory is null. I changed the code to this:

`const selectedIndex = inv.curSelectedItemIndex;`
`if (selectedIndex < 0) {...}`

`const selectedItem = inv.inventory[selectedIndex];`
`if (!selectedItem || !this.validMaterials.has(selectedItem.name) {...}`

I should mention this error only happened after I removed async from the method.

AndrewVieira: Async isn't related. I got the same error with the furnace when both tin and copper disappeared at the same time.

src/scenes/level.js
AndrewVieira: I kind of did what you asked, but I'm not sure if you like my solution.
* (In most recent push)

CurtisGreen (changes requested):
Code looking good.

I just tested it and noticed two things
1) It's not very intuitive to select bars in inventory -> right click anvil -> choose type -> left click to forge. Should we prompt the window on what to forge every time or keep it as is? Or if there's a better solution I'm all ears
2) Need to update `panels/skill-descriptions` to have the levels in which you can forge these new things

AndrewVieira:
One possible solution to reduce clicks is to auto-detect bars in the player's inventory if no bars are selected. The downside is that if there are multiple bar types, the type of bar auto-detected may not be the one the player wants to smith with. The player could still select the bar they want to smith with if they want.

Another solution could be to auto detect what bars the player has, then have multiple tabs for each bar type with the possible recipes shown for that bar when that tab is clicked. The downside to this is it would be a lot more effort to program and it differs much from the original Runescape interface.

CurtisGreen:
Yeah those could be good but they're a lot to add. I think we should keep it as-is for now and do an update later after some brainstorming

So just add the skill descriptions and I think it's good. Will have to see if Ethan finds anything else

src/targets/anvil.js
```diff
@@ -94,7 +134,9 @@ export class Anvil extends ClickableObject {
                 // Log click for stats
                 this.scene.stats.updateClickedTargetStat();
                 characterData.addSkillXp("smithing", item.xp);
-                this.scene.enemyKilled(item.questName);
+                if (this.scene.questExists(item.questName)) {
```
EthanVieira: Do we need this check here if we are also checking inside enemyKilled?

src/test-smithing-data.js
```diff
@@ -0,0 +1,121 @@
...
+    //hasCookies: false, // todo: maybe use this for tutorial?
+    name: "You",
+    gold: 0,
+    currentLevel: "TUTORIAL_ISLAND",
+    timesClicked: 0,
+    damageByClicking: 0,
+    damageByAutoClick: 0,
+    numberOfAutoClickers: 0,
+    questPoints: 0,
+    inventory: [
```
EthanVieira: Might want a large amount of ore as well if we think this smithing test data encompasses smelting as well. I'd also give a generous amount of gold just because why not :)

src/ui/modals/smithing-modal-window.js
EthanVieira: I don't think I have a preference. Combining the terms into one offset does help readability though

src/ui/modals/smithing-modal-window.js
```diff
@@ -0,0 +1,147 @@
...
+    setVisible(isVisible = true) {
+        this.visible = isVisible;
+        this.window.visible = isVisible;
+
+        for (let element of this.elements) {
+            element.visible = isVisible;
+        }
+    }
+
+    getChoice() {
```
EthanVieira: Don't think we need to override this method?

src/test-smithing-data.js
```diff
@@ -0,0 +1,121 @@
+export const testSmithingData = {
```
EthanVieira: I think we should create a directory for data files

EthanVieira (changes requested):
Looking really good!

One thing I would add is upon clicking a smithing option that you are unable to make, have it log the reason why you cannot make it. They can already see the number of bars required, but not the smithing level required.

src/test-smithing-data.js
CurtisGreen: 💰

AndrewVieira:
I fixed the issues you guys mentioned. For the creation of icons I did it the way the Curtis mentioned, but it's a little more complicated as you have to account for the window position and limit of icons per row.

I think I don't want to add the remaining bronze items for the time being. I'm not sure if we will even use all the item types, and I kind of want to start working on a different features.

Let me know if you guys find any more issues. Hopefully we can merge this code soon.

src/data/test-smithing-data.js
```diff
@@ -16,8 +16,10 @@ export const testSmithingData = {
         { item: "BronzePickaxe", count: 1 },
         { item: "Knife", count: 1 },
         { item: "Hammer", count: 1 },
-        { item: "Coin", count: 25 },
+        { item: "Coin", count: 69420 },
```
EthanVieira, line 19: 🥇

src/ui/modals/smithing-modal-window.js
```diff
@@ -62,48 +61,73 @@ export class SmithingModalWindow extends ModalWindow {
...
-            if (rowIndex >= this.maxIconsPerRow) {
-                positionX = this.x + this.leftOffset + this.horizontalPadding;
-                positionY += this.iconHeight + this.verticalPadding + 24;
-                rowIndex = 0;
-            }
+            const positionX = offsetX + padX * (index % this.maxIconsPerRow);
+            const positionY = offsetY + padY * parseInt(index / this.maxIconsPerRow);
 
             const item = await getItemClass(recipe, this.scene.window);
+            console.log(item.name);
```
EthanVieira: Is this leftover?

EthanVieira (commented):
Looks like maybe one leftover log statement, but besides that it all looks good to me!

src/ui/modals/smithing-modal-window.js
AndrewVieira: Oops I guess so, I'll wait for Curtis before updating

CurtisGreen: Wait on me for what?

CurtisGreen: I think we've done plenty on this PR already so anything else can probably be done later if needed

CurtisGreen (dismissed):

src/ui/modals/smithing-modal-window.js
AndrewVieira: I meant I was going to wait to see if you had any more changes in mind. I'll remove the console.log after I get off work today.

CurtisGreen (approved):

EthanVieira (approved):

PR #71, opened 2022-07-28
Improve autoclicker performance

EthanVieira:
* Allow only one object for each unique autoclicker in the list of active clan members
* Associate each unique clan member with the number owned
* Set dps to autoclicker.dps * number owned
* Add test data for testing autoclicker performance. Currently only includes 100 Bots. We will need to revisit when we have more autoclickers

### DIFF
--- a/src/cookie-io.js
+++ b/src/cookie-io.js
@@ -1,11 +1,10 @@
 import { defaultData } from "./data/default-data.js";
-import { testAutoclickerPerformanceData } from "./data/test-autoclicker-performance-data.js";
 import { CONSTANTS, FONTS } from "./constants/constants.js";
 import * as Utilities from "./utilities.js";
 
 function getDefaultData() {
     // Reset data (deep copy)
-    return JSON.parse(JSON.stringify(testAutoclickerPerformanceData));
+    return JSON.parse(JSON.stringify(defaultData));
 }
 
 class CharacterData {
--- a/src/cookie-io.js
+++ b/src/cookie-io.js
@@ -99,11 +99,11 @@ class CharacterData {
         this.characterData.clan.name = name;
     }
 
-    addClanMember(obj) {
-        if (obj in this.characterData.clan.members) {
-            this.characterData.clan.members[obj]++;
+    addClanMember(memberName) {
+        if (memberName in this.characterData.clan.members) {
+            this.characterData.clan.members[memberName]++;
         } else {
-            this.characterData.clan.members[obj] = 1;
+            this.characterData.clan.members[memberName] = 1;
         }
     }
     getClanMembers() {
--- a/src/cookie-io.js
+++ b/src/cookie-io.js
@@ -1,10 +1,11 @@
 import { defaultData } from "./data/default-data.js";
+import { testAutoclickerPerformanceData } from "./data/test-autoclicker-performance-data.js";
 import { CONSTANTS, FONTS } from "./constants/constants.js";
 import * as Utilities from "./utilities.js";
 
 function getDefaultData() {
     // Reset data (deep copy)
-    return JSON.parse(JSON.stringify(defaultData));
+    return JSON.parse(JSON.stringify(testAutoclickerPerformanceData));
 }
 
 class CharacterData {
@@ -100,7 +101,11 @@ class CharacterData {
     }
 
     addClanMember(obj) {
-        this.characterData.clan.members.push(obj);
+        if (obj in this.characterData.clan.members) {
+            this.characterData.clan.members[obj]++;
+        } else {
+            this.characterData.clan.members[obj] = 1;
+        }
     }
     getClanMembers() {
         return this.characterData.clan.members;
--- a/src/data/test-smithing-data.js
+++ b/src/data/test-smithing-data.js
@@ -6,7 +6,6 @@ export const testSmithingData = {
     timesClicked: 0,
     damageByClicking: 0,
     damageByAutoClick: 0,
-    numberOfAutoClickers: 0,
     questPoints: 0,
     inventory: [
         { item: "BronzeDagger", count: 1 },
@@ -26,8 +25,9 @@ export const testSmithingData = {
     },
     clan: {
         name: "The Black Knights",
-        members: [],
+        members: {},
     },
+    attackStyle: 0,
     // Skill XP
     skills: {
         attack: 0,
--- a/src/ui/panels/clan.js
+++ b/src/ui/panels/clan.js
@@ -94,7 +94,7 @@ export class Clan {
                 false,
                 startX,
                 startY + (this.clanMembers.length + 1) * yDiff
-            ); //TODO: add number to text
+            );
             member.setVisible(false);
             member.start(this.dashboard.currentScene);
             this.clanMembers.push(member);
@@ -129,7 +129,7 @@ export class Clan {
                     false,
                     startX,
                     startY + this.clanMembers.length * yDiff
-                ); //TODO: add number to text
+                );
                 this.clanMembers.push(member);
             }
 
--- a/src/ui/panels/clan.js
+++ b/src/ui/panels/clan.js
@@ -88,7 +88,7 @@ export class Clan {
         this.dashboard.currentScene.stats.resetAutoclickerDps();
         let savedClanMembers = characterData.getClanMembers();
         for (let memberName in savedClanMembers) {
-            let member = await getAutoclickerClass(memberName, this.scrollWindow);
+            const member = await getAutoclickerClass(memberName, this.scrollWindow);
             member.numberOwned = savedClanMembers[memberName];
             member.createText(
                 false,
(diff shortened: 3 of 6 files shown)


CurtisGreen:
Noice, I’ll review more thoroughly later but one thing I thought from this is we probably should use the in game update/tick rate instead of the timer that way we can remove the logic that creates and destroys that timer, whatcha think?

src/ui/panels/clan.js
```diff
@@ -99,20 +103,36 @@ export class Clan {
...
+            // Add new clan entry if new member else increment existing member by one
+            const index = this.clanMembers.findIndex((member) => {
+                if (member.name === memberName) {
+                    return true;
+                }
+                return false;
+            });
+
+            if (index != -1) {
+                this.clanMembers[index].numberOwned++;
```
CurtisGreen:
```js
const existingMember = this.clanMembers.find((member) => member.name === memberName);
if (existingMember) {
  existingMember.numberOwned++;
}
```

src/ui/panels/clan.js
```diff
@@ -99,20 +103,36 @@ export class Clan {
...
+            } else {
+                let member = getAutoclickerClass(memberName, this.scrollWindow);
```
CurtisGreen: `const newMember`

src/data/test-smithing-data.js
```diff
@@ -26,8 +25,9 @@ export const testSmithingData = {
     },
     clan: {
         name: "The Black Knights",
-        members: [],
+        members: {},
     },
+    attackStyle: 0,
```
CurtisGreen, line 30: Nice catch

AndrewVieira:
Sorry I haven't reviewed yet. I've been quite busy. I should be able to review tomorrow.

src/ui/panels/clan.js
```diff
@@ -99,20 +103,36 @@ export class Clan {
...
+            const index = this.clanMembers.findIndex((member) => {
+                if (member.name === memberName) {
+                    return true;
+                }
+                return false;
+            });
+
+            if (index != -1) {
+                this.clanMembers[index].numberOwned++;
+            } else {
```
AndrewVieira: It's not necessary, but if you use Curtis's revision, you could inverse the ```if (existingMember)``` condition to remove the else statement and two curly braces to make the code a little more readable. You would need to assign the newMember variable to the existingMember if ```(!existingMember)``` is true since we would always be incrementing existingMember in this case.

Example:

```
let existingMember = this.clanMembers.find((member) => member.name === memberName);

if (!existingMember) {
  const newMember = getAutoclickerClass(memberName, this.scrollWindow);
  newMember.createText(
                  false,
                  startX,
                  startY + this.clanMembers.length * yDiff
              );
  newMember.numberOwned = 0;
  this.clanMembers.push(newMember);
  existingMember = newMember;
} // No else statement needed!!!

existingMember.numberOwned++;

**Review by AndrewVieira — DISMISSED** [review]
I think Curtis's suggestions are good. I don't have much to add.

I was going to suggest that for the ```members: {},``` we use a Javascript map instead, but it seems that JSON.stringify doesn't work with maps by default.

File: src/ui/panels/clan.js (hunk 4 above)
Comment (CurtisGreen) [reply src/ui/panels/clan.js]: > I was going to suggest that for the members: {}, we use a Javascript map instead, but it seems that JSON.stringify doesn't work with maps by default.

I'm not sure what you mean like this, aren't JSON and JS maps/dictionaries are basically identical?

File: src/cookie-io.js [hunk 5]
```diff
@@ -100,7 +100,11 @@ class CharacterData {
     }
 
     addClanMember(obj) {
-        this.characterData.clan.members.push(obj);
+        if (obj in this.characterData.clan.members) {
```
CurtisGreen: Didn't know you could use `in` like in python, nice. Should rename `obj` -> `key` or `memberName` to be more accurate now that it's not an array

src/ui/panels/clan.js
CurtisGreen: Also since we're changing the save data for members to be to be a dictionary maybe we should do the same for the game objects too

src/ui/panels/clan.js
AndrewVieira: Kind of, but they're a little different. Both maps and objects use key and value pairs to store data.

The one big difference that might be useful are that maps keep the order of elements preserved, while objects do not.

I also prefer the syntax of maps over objects in this situation. For example to determine if a map has a key, you can just do '''map.has(key)''', while for object the equivalent would be '''object.key === undefined'''. Another example is '''map.size''' vs '''Object.keys(object).length''' to get the number of items.

We should use object though, since JSON supports objects and not maps. It sounds like there may be future support for maps in the future though.

I was going to put an article showcasing the differences but it's a pain to post links on my phone.

CurtisGreen: Oh gotcha, I guess I never realized they had separate object/dictionaries and an actual Map class. Good to know

src/cookie-io.js
EthanVieira: https://github.com/EthanVieira/rs-clicker/pull/71/commits/8e37d85adae8b81c4d9a1d9b3e40c7f746834278

src/ui/panels/clan.js
EthanVieira: https://github.com/EthanVieira/rs-clicker/pull/71/commits/8e37d85adae8b81c4d9a1d9b3e40c7f746834278

src/ui/panels/clan.js
EthanVieira: Decided not to change that to a dictionary as I don't currently see a need for it. If we ever need to retrieve them by name then we can change it

EthanVieira:
> Noice, I’ll review more thoroughly later but one thing I thought from this is we probably should use the in game update/tick rate instead of the timer that way we can remove the logic that creates and destroys that timer, whatcha think?

I created a trello task for this :)

src/ui/panels/clan.js
EthanVieira: Decided not to change to a dict for now. If it is more convenient later on then we can update

src/ui/panels/clan.js
```diff
@@ -87,10 +87,14 @@ export class Clan {
         // Reset dps counter before refreshing autoclickers
         this.dashboard.currentScene.stats.resetAutoclickerDps();
         let savedClanMembers = characterData.getClanMembers();
-        for (let index = 0; index < savedClanMembers.length; index++) {
-            let memberName = savedClanMembers[index];
+        for (let memberName in savedClanMembers) {
             let member = await getAutoclickerClass(memberName, this.scrollWindow);
```
CurtisGreen: I might've mixed up my comments, this one should be a const, the other one either needs to be a `let` or you need to create a new variable

src/ui/panels/clan.js
```diff
@@ -99,20 +103,28 @@ export class Clan {
...
                 yDiff = 18;
 
-            // Add text to the list
-            let index = this.clanMembers.length;
-            member.createText(false, startX, startY + index * yDiff);
-            this.clanMembers.push(member);
+            // Add new clan entry if new member else increment existing members
+            const member = this.clanMembers.find((member) => member.name === memberName);
+            if (!member) {
+                member = getAutoclickerClass(memberName, this.scrollWindow);
```
CurtisGreen, line 119: Declared as const but then it's overwritten, see my previous comment. This either needs to be changed to be
```js
const existingMember = this.clanMembers.find((member) => member.name === memberName);
if (!existingMember ) {
  const newMember = ...
  ...
  newMember.numberOwned = 1;
}
else {
  existingMember .numberOwned++;
}
```

or

```js
let member = this.clanMembers.find((member) => member.name === memberName);
if (!member ) {
  member = ...
  ...
}
member .numberOwned++;
```

CurtisGreen (dismissed):

src/ui/panels/clan.js
```diff
@@ -99,20 +103,28 @@ export class Clan {
...
+                member.createText(
+                    false,
+                    startX,
+                    startY + this.clanMembers.length * yDiff
+                );
+                this.clanMembers.push(newMember);
```
AndrewVieira: This is suppose to be ```this.clanMembers.push(member);``` instead of ```newMember``` right? This is a bit weird. I ran the code and it works fine, but newMember variable isn't declared anywhere in the code before. I'm not sure why no error is popping up. (I used a private window to reset cookies.)

AndrewVieira (commented):
I thought I would catch a bug, but the program runs normally. Does anyone know why? (Listed below.)

src/ui/panels/clan.js
CurtisGreen: Yeah that's kinda crazy, just javascript things I guess?

src/ui/panels/clan.js
EthanVieira: https://github.com/EthanVieira/rs-clicker/pull/71/commits/0828d5299ae78d1c416ad0dfa5957d3c29402087

src/ui/panels/clan.js
EthanVieira: whoops 🤷‍♂️ https://github.com/EthanVieira/rs-clicker/pull/71/commits/0828d5299ae78d1c416ad0dfa5957d3c29402087

AndrewVieira (commented):
Code looks fine to me. I ran the code and it seems to be working fine.

PR #72, opened 2022-08-05
Small test refactor

EthanVieira:
* Made the test data files more explicit in what they are changing
* Fixed a bug where the total number of enemies killed stat would not increment if all quests were completed for the enemy you are killing


EthanVieira:
OBE. Opening PR for larger set of changes

PR #73, opened 2022-08-12
Dynamic data loading

EthanVieira:
* Fixed a bug where the total number of enemies killed stat would not increment if all quests were completed for the enemy you are killing
* Moved the test data into one file and made them more explicit in what they are changing
* Added command parsing to chat
* Added load command for loading test data dynamically
Usage: /load name-of-test-data


src/data/test-data.js
```diff
@@ -0,0 +1,27 @@
+import { defaultData } from "./default-data.js";
+// ---------------------------------------------------------------------
+let allLevelsUnlockedData = JSON.parse(JSON.stringify(defaultData));
```
AndrewVieira: ```const allLevelsUnlockedData``` instead of ```let```. I think you can make JSON objects const and still change their member variables.

src/data/test-data.js
```diff
@@ -0,0 +1,27 @@
...
+for (let scene in allLevelsUnlockedData["levels"]) {
+    allLevelsUnlockedData["levels"][scene]["questCompleted"] = true;
+    allLevelsUnlockedData["levels"][scene]["unlocked"] = true;
+}
+// ---------------------------------------------------------------------
+let autoclickerPerformanceTestData = JSON.parse(JSON.stringify(allLevelsUnlockedData));
+
+autoclickerPerformanceTestData["clan"]["members"]["Bot"] = 100;
+// ---------------------------------------------------------------------
+let smithingTestData = JSON.parse(JSON.stringify(allLevelsUnlockedData));
```
AndrewVieira: ```const```

src/data/test-data.js
```diff
@@ -0,0 +1,27 @@
+import { defaultData } from "./default-data.js";
+// ---------------------------------------------------------------------
+let allLevelsUnlockedData = JSON.parse(JSON.stringify(defaultData));
+
+for (let scene in allLevelsUnlockedData["levels"]) {
+    allLevelsUnlockedData["levels"][scene]["questCompleted"] = true;
+    allLevelsUnlockedData["levels"][scene]["unlocked"] = true;
+}
+// ---------------------------------------------------------------------
+let autoclickerPerformanceTestData = JSON.parse(JSON.stringify(allLevelsUnlockedData));
```
AndrewVieira: ```const```

src/ui/chat-window.js
```diff
@@ -454,18 +455,65 @@ export class ChatScene extends Phaser.Scene {
...
+                        0,
+                        this.userMessage.text.length - 1
                     );
+
+                    // check for commands
+                    if (this.userMessage.text[0] == "/") {
+                        let words = this.userMessage.text.split(" ");
+
+                        // remove "/" from command word
+                        let command = words[0].substr(1).toLowerCase();
```
AndrewVieira: ```const```

src/ui/chat-window.js
```diff
@@ -454,18 +455,65 @@ export class ChatScene extends Phaser.Scene {
...
-                        }
+                    // remove prompt character
+                    this.userMessage.text = this.userMessage.text.substr(
+                        0,
+                        this.userMessage.text.length - 1
                     );
+
+                    // check for commands
+                    if (this.userMessage.text[0] == "/") {
+                        let words = this.userMessage.text.split(" ");
```
AndrewVieira: ```const```

AndrewVieira (commented):
Looks interesting! I tried to use the command parsing with the /load command but chat kept repeating the message and nothing happened. I was on Linux and tested with both Firefox and Chromium. I'll test on Windows later.

src/data/test-data.js
EthanVieira: Added all these consts here: https://github.com/EthanVieira/rs-clicker/pull/73/commits/0483fe8361308e7b3ba7f08ab2a5eee54601ef26

EthanVieira:
> Looks interesting! I tried to use the command parsing with the /load command but chat kept repeating the message and nothing happened. I was on Linux and tested with both Firefox and Chromium. I'll test on Windows later.

None of this should be OS dependent. Are you sure you had the branch checked out and up to date? It shouldn't treat any text as a normal message if it starts with a forward slash

AndrewVieira:
I did git status and git pull again to double check, but I'm in the right branch "dynamic-data-loading". Maybe I'm looking in the wrong place? I click on the green log button near the bottom left and type my command there. Is that correct? Is there another chat?

EthanVieira:
Just in the normal chat box. You should see Loading: data-name in the chat when it works. Otherwise, if you start with forward slash and it is an incorrect command you will get an error message.

image

AndrewVieira:
I got the load command to work on windows. I'm not sure what's up with my Linux environment. There is an issue with Firefox in that you can't type dashes "-", but everything works fine in Google Chrome.

AndrewVieira:
@EthanVieira I could look into the Firefox issue to see if what's up with the dashes if you are fine with that. I would do it on this branch.

EthanVieira:
> @EthanVieira I could look into the Firefox issue to see if what's up with the dashes if you are fine with that. I would do it on this branch.

You can if you'd like. Ultimately I don't think it's that huge of a deal as long as those who are testing it are using chrome.

EthanVieira:
It has been nearly three weeks, so I'm merging.

PR #76, opened 2022-09-05
add set-level and set-xp commands

EthanVieira:
image

* There is a future task to resize the xp box to hold the long xp numbers [here](https://trello.com/c/Kfo5PeEh/166-resize-skill-xp-box-on-skill-hover-to-fit-the-length-of-the-numbers-shown)
* Usage:
            /set-level/set-xp name-of-skill level/xp
            if name-of-skill is "all", set all skills to the given level/xp
            if level/xp is "max", set the skill/xp to 99/200,000,000
            e.g.
            /set-level/set-xp fletching 72
            /set-level/set-xp all max

### DIFF
--- a/src/constants/constants.js
+++ b/src/constants/constants.js
@@ -66,6 +66,11 @@ export const CONSTANTS = {
         CLAN: "CLAN",
         ATTACK_STYLE: "ATTACK_STYLE",
     },
+    LIMITS: {
+        MAX_LEVEL: 99,
+        MAX_XP: 200_000_000,
+        XP_FOR_99: 13_034_431,
+    },
     UTILS: {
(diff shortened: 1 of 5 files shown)


src/constants/constants.js
```diff
@@ -66,6 +66,11 @@ export const CONSTANTS = {
         CLAN: "CLAN",
         ATTACK_STYLE: "ATTACK_STYLE",
     },
+    LIMITS: {
+        MAX_LEVEL: 99,
+        MAX_XP: 200_000_000,
```
CurtisGreen, line 71: Does this work with the underscores?

CurtisGreen: Woah it does, weird

CurtisGreen (approved):

Issue #77, opened 2022-09-15
typing in logs


jeff092111:
when ever you type it normally does the function of adding the letter twice

CurtisGreen:
What browser are you using? (Chrome, Firefox, safari, etc)

EthanVieira:
I noticed that keys often get duplicated when pressing multiple different keys at the same time. It happens frequently in both Chrome and Firefox. Investigating it now

EthanVieira:
Seems to be a Phaser issue. Try pressing "asdf" at the same time here: https://phaser.io/examples/v3/view/input/keyboard/text-entry.

Put in a hack to keep track of recently used letters. Duplicating letters by holding a key down is a bit slower, but it is necessary to fix this. PR is here: https://github.com/EthanVieira/rs-clicker/pull/78

PR #78, opened 2022-09-21
Duplicate letters

EthanVieira:
* Fixed a bug where letters are duplicated when pressing multiple keys simultaneously (or near simultaneously)
* Issue: https://github.com/EthanVieira/rs-clicker/issues/77

### DIFF
--- a/src/ui/chat-window.js
+++ b/src/ui/chat-window.js
@@ -513,8 +513,8 @@ export class ChatScene extends Phaser.Scene {
                     // Phaser has an input bug where pressing multiple keys simultaneously
                     // (or near-simultaneously) results in duplicate letters being detected as input.
                     // Here we add recently typd letters to a decaying list so that they aren't duplicated
-                    // within 100 milliseconds
-                    this.recentKeys.put(event.keyCode, 100);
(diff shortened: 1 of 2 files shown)


src/ui/chat-window.js
```diff
@@ -497,14 +498,23 @@ export class ChatScene extends Phaser.Scene {
...
                             this.userMessage.text.length - 1
                         ) +
                         event.key +
                         "*";
+
+                    // Phaser has an input bug where pressing multiple keys simultaneously
+                    // (or near-simultaneously) results in duplicate letters being detected as input.
+                    // Here we add recently typd letters to a decaying list so that they aren't duplicated
+                    // within 100 milliseconds
+                    this.recentKeys.put(event.keyCode, 100);
```
CurtisGreen:
```suggestion
                    this.recentKeys.put(event.keyCode, 10);
```
Played around with it and I think we can lower it a lot and basically have no drawbacks as long as it's non-zero

EthanVieira: Ah nice nice. Updated!

CurtisGreen (approved):
Nice solution

PR #79, opened 2022-09-21
add add-item command

EthanVieira:
image
* Usage:
  /add-item name-of-item amount
  amount is optional and defaults to 1
  if amount is "max", set the amount to 2,147,483,647 if stackable or 28 if not stackable
  e.g.
 /add-item gold max
 /add-item BronzeDagger

### DIFF
--- a/src/constants/constants.js
+++ b/src/constants/constants.js
@@ -70,6 +70,8 @@ export const CONSTANTS = {
         MAX_LEVEL: 99,
         MAX_XP: 200_000_000,
         XP_FOR_99: 13_034_431,
+        MAX_ITEM_STACK: 2_147_483_647,
+        MAX_INVENTORY_SPACE: 28,
     },
     UTILS: {
         MILLIS_IN_YEAR: 365 * 24 * 60 * 60 * 1000,
--- a/src/utilities.js
+++ b/src/utilities.js
@@ -37,6 +37,7 @@ export async function getItemClass(itemName, scene) {
         let itemClass = itemClasses[itemName];
         return new itemClass.default(scene);
     } else {
+        console.log("Invalid item name in getItemClass - returning null.");
         return null;
     }
 }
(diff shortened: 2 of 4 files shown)


src/constants/constants.js
```diff
@@ -70,6 +70,8 @@ export const CONSTANTS = {
         MAX_LEVEL: 99,
         MAX_XP: 200_000_000,
         XP_FOR_99: 13_034_431,
+        MAX_ITEM_STACK: 2_147_483_647,
```
CurtisGreen, line 73: lol

src/utilities.js
```diff
@@ -33,9 +33,12 @@ export function calcXpForLevel(lv) {
...
 var itemClasses = {};
 export async function getItemClass(itemName, scene) {
-    let itemClass = itemClasses[itemName];
-
-    return new itemClass.default(scene);
+    if (itemClasses.hasOwnProperty(itemName)) {
+        let itemClass = itemClasses[itemName];
+        return new itemClass.default(scene);
+    } else {
+        return null;
```
CurtisGreen, line 41: I think we should log if we're passing in a bad item name here

CurtisGreen (commented):
I thought the INT max would be a problem but it handles it like a champ lol

src/utilities.js
EthanVieira: Updated! https://github.com/EthanVieira/rs-clicker/pull/79/commits/7291bf8df46a141986148de968f0b8284f32358a

src/utilities.js
```diff
@@ -33,9 +33,13 @@ export function calcXpForLevel(lv) {
...
 var itemClasses = {};
 export async function getItemClass(itemName, scene) {
-    let itemClass = itemClasses[itemName];
-
-    return new itemClass.default(scene);
+    if (itemClasses.hasOwnProperty(itemName)) {
+        let itemClass = itemClasses[itemName];
+        return new itemClass.default(scene);
+    } else {
+        console.log("Invalid item name in getItemClass - returning null.");
```
CurtisGreen:
```suggestion
        console.log("Invalid item name in getItemClass - returning null.", itemName);
```

CurtisGreen (approved):

PR #80, opened 2022-09-24
Add member command

EthanVieira:
* Usage:
   /add-member name-of-member amount
   amount is optional and defaults to 1
   e.g.
   /add-member bot 10

* added two new trello tasks: 
   * Improve autoclicker performance even more
   * Don't let chatbox close from a pointerout event on an autoclicker


CurtisGreen (approved):
You have perfected adding commands

PR #81, opened 2022-09-25
resize skill xp box on hover

EthanVieira:


CurtisGreen:
Looks good, while you're at it maybe set height based on lines of text? Would fix the problem with total skill level having an empty line at the bottom
https://i.imgur.com/Zrmrerq.png

AndrewVieira (commented):
I noticed a bug. The tool tip with the level and experience will be partially hidden by the text box.

CurtisGreen:
> I noticed a bug. The tool tip with the level and experience will be partially hidden by the text box.

Yeah that's been around for a bit, need to reorder the scenes so that text-box is loaded before dashboard. If there's no collateral then it's an easy change. If it affects other things it's pretty tough

EthanVieira:
> Looks good, while you're at it maybe set height based on lines of text? Would fix the problem with total skill level having an empty line at the bottom https://i.imgur.com/Zrmrerq.png

Updated! https://github.com/EthanVieira/rs-clicker/pull/81/commits/6f3f19474063540eba4a20b72982b90ddc6555e4

EthanVieira:
> > I noticed a bug. The tool tip with the level and experience will be partially hidden by the text box.
> 
> Yeah that's been around for a bit, need to reorder the scenes so that text-box is loaded before dashboard. If there's no collateral then it's an easy change. If it affects other things it's pretty tough

Fixed and left some notes about it https://github.com/EthanVieira/rs-clicker/pull/81/commits/6f3f19474063540eba4a20b72982b90ddc6555e4

AndrewVieira (approved):
Bug fixed. All is guud.

PR #82, opened 2022-10-04
Three small changes

EthanVieira:
* clan members only show examine when clicked
 * up key types previously entered message
 * enter opens and closes chat
     * Currently allowed to open during shop scene. We can discuss this

### DIFF
--- a/src/ui/chat-window.js
+++ b/src/ui/chat-window.js
@@ -1,4 +1,4 @@
-import { OBJECT_TYPES, CONSTANTS, FONTS } from "../constants/constants.js";
+import { OBJECT_TYPES, CONSTANTS, FONTS, KEY_CODES } from "../constants/constants.js";
 import { ScrollWindow } from "./scroll-window.js";
 import { TextRow } from "./text-row.js";
 import { Button } from "./button.js";
@@ -389,8 +389,7 @@ export class ChatScene extends Phaser.Scene {
             .setDepth(9);
 
         let keyboardInput = this.input.keyboard.on("keydown", (event) => {
-            // enter
-            if (event.keyCode == 13) {
+            if (event.keyCode == KEY_CODES.ENTER) {
                 let inputValue = parseInt(
                     promptInput.text.substr(0, promptInput.text.length - 1)
                 );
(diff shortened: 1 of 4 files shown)


src/ui/chat-window.js
```diff
@@ -476,26 +475,42 @@ export class ChatScene extends Phaser.Scene {
...
                 this.userMessage.text =
                     this.userMessage.text.substr(0, this.userMessage.text.length - 2) +
                     "*";
             }
 
+            if (event.keyCode == Phaser.Input.Keyboard.KeyCodes.UP) {
+                this.userMessage.text = this.previousMessage;
+            }
+
+            // There are some firefox only keycodes that we might want to look into
```
CurtisGreen, line 496: Like what?

src/ui/chat-window.js
```diff
@@ -453,9 +453,8 @@ export class ChatScene extends Phaser.Scene {
         }
 
         this.chatInput = this.input.keyboard.on("keydown", (event) => {
-            // enter
-            if (event.keyCode == 13) {
-                if (this.userMessage.text != "") {
+            if (event.keyCode == Phaser.Input.Keyboard.KeyCodes.ENTER) {
```
CurtisGreen: Object string is pretty long, maybe pre-declare the keycodes so it's cleaner and can be used for all of the next ones

```js
const KeyCodes = Phaser.Input.Keyboard.KeyCodes;
if (event.KeyCode == KeyCodes.ENTER)
    // ...
```

src/ui/chat-window.js
EthanVieira: At the bottom of this file: https://github.com/photonstorm/phaser/blob/v3.51.0/src/input/keyboard/keys/KeyCodes.js

src/ui/chat-window.js
EthanVieira: Updated!

CurtisGreen (approved):

src/ui/chat-window.js
CurtisGreen: Gotcha, can add them later if necessary but right now I don't think it's a big deal

PR #84, opened 2022-10-16
add equip and unequip sfx

EthanVieira:


CurtisGreen (approved):
Bow & staff are pretty hard to hear compared to the others. Do you think we should bump those up or is that expected

EthanVieira:
> 
Expected!

EthanVieira:
> Bow & staff are pretty hard to hear compared to the others. Do you think we should bump those up or is that expected

Expected!

PR #86, opened 2023-04-27
add fishing skill and draynor fishing scene

EthanVieira:
* Started the Fishing skill
* Added Draynor Fishing scene 
* Added shrimps/anchovies and small fishing net
* Added unknown land song
* Migrated mp3s to oggs
* Fixed a bug where skill text hover was behind chat window
* Reduced font size for quests
* Added fishing test data
* Added support for unique drops
* Discussion points for Fishing skill - 
  Currently:
     - the targets that count for quests are the types of fishing spots
     - xp is based on clicks required and not fish gained
     - fish are considered "unique" drops so you can only obtain one potential drop. If the sum of potential drop rates is < 100% then no drop is possible despite gaining xp.

image


src/targets/resources/fishing-spots/fishing-spot.js
```diff
@@ -0,0 +1,17 @@
+import { Resource } from "../resource.js";
+import { CONSTANTS } from "../../../constants/constants.js";
+import { calcLevel } from "../../../utilities.js";
+import { characterData } from "../../../cookie-io.js";
```
CurtisGreen: Looks like several of these imports are unused

src/targets/resources/fishing-spots/small-net-fishing-spot.js
```diff
@@ -0,0 +1,44 @@
...
+            varName: "smallNetFishingSpot",
+            images: [
+                {
+                    name: "fishing-spot",
+                    path: "src/assets/sprites/FishingSpot.png",
+                    scale: 1,
+                },
+            ],
+            neededClicks: 5,
+            uniqueDrops: [
```
CurtisGreen, line 24: I don't think I understand the difference between drops & unique drops

src/targets/target.js
```diff
@@ -84,6 +88,41 @@ export class Target extends ClickableObject {
...
+                            return false;
+                        }
+                    }
+                    //continue
+                    return true;
+                });
+            }
+
+            // General drops
+            // Can obtain multiple
```
CurtisGreen, line 121: Nvm I get it

src/targets/target.js
```diff
@@ -84,6 +88,41 @@ export class Target extends ClickableObject {
...
+            // Separate rolls for unique and general
+
+            // Unique drops
+            // Can only obtain one
+            if (this.uniqueDrops) {
+                const randomNum = Math.random();
+                let threshold = 0;
+
+                this.uniqueDrops.every((item) => {
+                    const canGetDrop = Object.keys(item.requiredLevels).every((skill) => {
```
CurtisGreen: Can replace with arrow function (if it looks nicer)

```js
.every((skill) => {
  return (
    calcLevel(characterData.getSkillXp(skill)) >=
    item.requiredLevels[skill]
  )
});
```
=>
```js
.every((skill) =>
  calcLevel(characterData.getSkillXp(skill)) >= item.requiredLevels[skill]
);
```

src/targets/target.js
```diff
@@ -84,6 +88,41 @@ export class Target extends ClickableObject {
...
             // Calculate item drops
+            // Separate rolls for unique and general
+
+            // Unique drops
+            // Can only obtain one
+            if (this.uniqueDrops) {
+                const randomNum = Math.random();
+                let threshold = 0;
+
+                this.uniqueDrops.every((item) => {
```
CurtisGreen: Since we're not using the result I think we should just use a `for of` loop here, and then you can use a normal `break` below instead of returning true/false

```js
for (const item of this.uniqueDrops) {


}
```

src/targets/target.js
```diff
@@ -84,6 +88,41 @@ export class Target extends ClickableObject {
...
+                            calcLevel(characterData.getSkillXp(skill)) >=
+                            item.requiredLevels[skill]
+                        );
+                    });
+
+                    if (canGetDrop) {
+                        // Assumes sum of rates <= 100%
+                        threshold += item.rate;
+                        if (threshold > randomNum) {
+                            let droppedItem = new item.item(this.scene.dashboard);
```
CurtisGreen: Because we never re-assign `droppedItem` it can be a const

src/ui/panels/quests.js
```diff
@@ -209,17 +209,23 @@ export class Quests {
...
                     let questText = this.questsScrollWindow.add
-                        .text(0, 0, prettyPrintCamelCase(enemy) + "s: ", {
-                            fill: isQuestComplete ? "#00ff00" : "yellow",
-                            font: "16px runescape",
-                        })
+                        .text(
+                            0,
+                            0,
+                            prettyPrintCamelCase(enemy) +
+                                (enemy.endsWith("s") ? ": " : "s: "),
```
CurtisGreen, line 216: Formatter made it look wonky but this is a good idea nice

src/ui/scroll-window.js
```diff
@@ -275,5 +275,14 @@ export class ScrollWindow extends Phaser.Scene {
             this.scrollBackground.visible = isVisible;
             this.scrollBar.visible = isVisible;
         }
+
+        if (isVisible) {
+            // Whenever we load the skills scene, we bring the dashboard to the top so
```
CurtisGreen, line 280: Why do you love paragraph comments so much lol

My humble suggestion:
```
// Prevent scrollbar from being hidden behind other scenes
```

CurtisGreen (commented):
I'll run it later today, overall looking pretty good 🎣

src/ui/scroll-window.js
EthanVieira: :) There is a very specific reason why we need to do this now, and if in the future we need to not have the scrollbar be on top, this comment will help us maintain the fix while accommodating the new. I will change it if you really really want but the context could potentially save a lot of time in the future

src/targets/target.js
EthanVieira: Hmm we might use the value in the future if we don't want to have a general drop when we receive a unique. I'll change for now

CurtisGreen (commented):
image

Minimap is off

EthanVieira:
> image
> 
> Minimap is off

CurtisGreen (approved):

EthanVieira:
[Fishing](https://trello.com/c/cXenBVE3/171-fishing)

PR #89, opened 2023-04-30
add defense/strength skill 

EthanVieira:
* added strength and defense 
* allow xp counter to show xp for multiple skills simultaneously
* discussion point: 
      - should we call it "defense"? or the UK version that is found in RS: "defence"
image


EthanVieira:
[Strength/Defence](https://trello.com/c/TgGe6HGb/170-strength-defence)

CurtisGreen (commented):
I think the UK version

src/cookie-io.js
```diff
@@ -163,37 +163,46 @@ class CharacterData {
...
-
-            // Play level up sfx
-            if (curLevel > prevLevel) {
-                const audioScene = this.getScene(CONSTANTS.SCENES.AUDIO);
-                audioScene.playSfx(skill + "-level-up");
+    // map of skill: xp as input
+    addSkillXp(skillXpMap) {
+        let animationMap = {};
+
+        for (var skill of Object.keys(skillXpMap)) {
```
CurtisGreen:
```suggestion
        for (const skill in skillXpMap
```

CurtisGreen: Or just replace the method input with an array of objects:

Ex: `[{skill: "defence", xp: "5"}, {skill: "ranged", xp: "2"}]`

src/items/equipment/equipment.js
```diff
@@ -82,7 +82,11 @@ export default class Equipment extends Item {
                     .playSfx(this.equipSound, false);
             } else {
                 console.log("Not high enough level to equip that.");
-                let skillText = getRequiredCombatSkill(this.skill);
+
+                const skillText =
+                    this.skill == EQUIPMENT.WEAPON_TYPES.MELEE
+                        ? "attack"
+                        : getRequiredCombatSkill(this.skill);
```
CurtisGreen, line 89: Since we pass in `this.skill` into the method we can move this ternary logic there

src/scenes/audio.js
```diff
@@ -44,13 +44,15 @@ export class AudioScene extends Phaser.Scene {
         const skills = [
             "Attack",
             "Cooking",
+            "defence",
```
CurtisGreen: Should be uppercase

src/targets/resources/resource.js
```diff
@@ -94,7 +94,9 @@ export class Resource extends Target {
     onClick(clickValue) {}
 
     onCompletion() {
-        characterData.addSkillXp(this.skill, this.neededClicks);
+        let skillXpMap = {};
+        skillXpMap[this.skill] = this.neededClicks;
+        characterData.addSkillXp(skillXpMap);
```
CurtisGreen:
```suggestion
        characterData.addSkillXp({ [this.skill]: this.neededClicks ));
```

src/ui/animation.js
```diff
@@ -21,16 +21,45 @@ export class Animation extends Phaser.Scene {
...
+        if (
+            !this.dashboard.xpCounterOn ||
+            skillXpMap === undefined ||
+            skillXpMap.length == 0
+        ) {
             return;
         }
 
-        const startX = 470;
+        let targets = [];
```
CurtisGreen: You can use const for arrays and objects as const only prevents them from being re-assigned but doesn't prevent them from being modified

src/ui/animation.js
```diff
@@ -45,50 +74,26 @@ export class Animation extends Phaser.Scene {
...
-                if (
-                    xpText.x >= endX - 1 &&
-                    xpText.x <= endX + 1 &&
-                    xpText.y >= endY - 1 &&
-                    xpText.y <= endY + 1
-                ) {
-                    xpText.destroy();
-                    skillIcon.destroy();
-                    graphics.destroy();
+                if (xpText.y >= endY - 1 && xpText.y <= endY + 1) {
```
CurtisGreen, line 95: Does removing the X axis check not mess up attack animations?

src/utilities.js
```diff
@@ -115,7 +116,17 @@ export function aOrAn(followingWord) {
 export function getRequiredCombatSkill(skillConstant) {
     switch (skillConstant) {
         case EQUIPMENT.WEAPON_TYPES.MELEE:
-            return "attack";
+            switch (characterData.getAttackStyle()) {
+                case 0:
```
CurtisGreen: If we don't have constants for these we should make some

CurtisGreen (changes requested):

src/targets/resources/resource.js
CurtisGreen:
```js
const key = "keyString";
const value = "valueString";
const obj = { [key]: value}; // { keyString: "valueString" }
```

src/items/equipment/equipment.js
EthanVieira: The ternary logic is different and not used for each usage of the method. It's a bit awkward and if it gets more complex in the future might look into refactoring the flow

src/cookie-io.js
EthanVieira: I think I prefer the way I have it but not totally convinced either way. It does clean up the for loop a little bit as you demonstrated, but I like the simpler interface with a single object as input `{defence: 5, ranged: 2}`.

src/targets/resources/resource.js
EthanVieira: Noice

src/ui/animation.js
EthanVieira: This is only for the xp animation

src/cookie-io.js
```diff
@@ -163,37 +163,46 @@ class CharacterData {
...
-
-            const curLevel = Utilities.calcLevel(this.characterData.skills[skill]);
-
-            // Play level up sfx
-            if (curLevel > prevLevel) {
-                const audioScene = this.getScene(CONSTANTS.SCENES.AUDIO);
-                audioScene.playSfx(skill + "-level-up");
+    // map of skill: xp as input
+    addSkillXp(skillXpMap) {
+        let animationMap = {};
```
CurtisGreen:
```suggestion
        const animationMap = {};
```

src/targets/enemies/enemy.js
```diff
@@ -265,15 +267,41 @@ export class Enemy extends Target {
...
+        return calcLevel(characterData.getSkillXp(requiredCombatSkill));
+    }
 
-        // Increase attack/ranged/magic XP
+    increaseXp(hitValue) {
+        const skill = this.scene.dashboard.equipment.equipment.WEAPON?.skill;
         const xpModifier = 1; // OSRS has an xp mod of 4 but that's assuming your attack speed is much lower
         let xpIncrease = xpModifier * hitValue;
-        characterData.addSkillXp(getRequiredCombatSkill(skill), xpIncrease);
+        let skillXpMap = {};
```
CurtisGreen:
```suggestion
        const skillXpMap = {};
```

CurtisGreen (approved):

PR #90, opened 2023-05-01
add all level-up jingles and add firemaking skill

EthanVieira:
* Add firemaking skill
* Add all level-up jingles


EthanVieira:
[Firemaking](https://trello.com/c/K4OXUrov/173-firemaking)

src/items/tools/tinderbox.js
```diff
@@ -0,0 +1,38 @@
...
+    scene;
+
+    constructor(scene) {
+        super();
+        this.scene = scene;
+    }
+
+    getRecipe(itemName) {
+        let output = {
+            className: "",
```
CurtisGreen: I think null might be better than checking for empty string

src/items/tools/tool.js
```diff
@@ -35,25 +35,35 @@ export default class Tool extends Item {
...
+                        "Fletched " + newItemName + ".";
+                    characterData.addSkillXp({ fletching: recipe.xpGiven });
+                }
+            } else {
+                // No new item is created
+                item.setNumItems(item.numItems - recipe.numRequiredItems);
                 outputString =
-                    "Crafted " + newItemName + ", added " + recipe.xpGiven + "xp";
-                characterData.addSkillXp({fletching: recipe.xpGiven});
+                    "Firemaked " + item.name + ".";
```
CurtisGreen: Since this method is so different from the fletching one I think we should refactor this a bit. A couple of options:

1. Override `craft` method in child class
2. Require child to pass in action callback function into the craft method to handle custom logic
3. Create sub classes under tools to handle things in common, like grouping tools that create items as one subclass

EthanVieira: Went with 1

src/items/tools/tinderbox.js
```diff
@@ -0,0 +1,67 @@
...
+    // Scenes
+    scene;
+
+    constructor(scene) {
+        super();
+        this.scene = scene;
+    }
+
+    getRecipe(itemName) {
+        let output = {
```
CurtisGreen:
```suggestion
        const output = {
```

src/items/tools/knife.js
```diff
@@ -37,4 +40,47 @@ export default class Knife extends Tool {
...
+        let outputString = "";
+        const recipe = this.getRecipe(item.name);
+        console.log("Combining", this.name, item.name);
+
+        // Fletch item if possible
+        if (item.numItems >= recipe.numRequiredItems) {
+            const dashboard = characterData.getScene(CONSTANTS.SCENES.DASHBOARD);
+
+            if (recipe.className) {
+                let newItem = await getItemClass(recipe.className, dashboard);
```
CurtisGreen:
```suggestion
                const newItem = await getItemClass(recipe.className, dashboard);
```

CurtisGreen (approved):

PR #91, opened 2023-07-30
Gnome Village Agility Course

EthanVieira:
* Add agility skill
* Add Gnome Village Agility Course
* Update World Map
* Discussion Points
     - Created trello tasks to resize right click menu width based on text and allow scrolling in/out on world map
         * Currently the log obstacle right click text is messed up because of the first one
     - Not sure if agility courses should cycle through the different obstacles. I chose to do a single obstacle for now because getting assets of the individual obstacles isn't as easy as it is to get assets of monsters. We could have different background images of the obstacles and cycle through them, but with high levels and auto clickers I think it would get way too wild. If we really want it then we'll have to get our paint skills going and separate the obstacles in screenshots by hand.


CurtisGreen:
Loading all levels doesn't work, looks like an issue with the quest amount not being found: 
image

image

EthanVieira:
Hmm are you sure you have all the changes? I saw that before I updated main.js, but doesn't look like I have anything not pushed and it works fine for me

CurtisGreen:
> Hmm are you sure you have all the changes? I saw that before I updated main.js, but doesn't look like I have anything not pushed and it works fine for me

Yeah I think it's because I didn't do it in an incognito tab there must be an issue with old cookies. I did find a different minor issue though, skill text needs a line break. Also I think the hitbox should be adjusted on the log, I was pressing the center and thought it didn't work but it only works if you press the bottom of the log
image

EthanVieira:
I was going for what it would be like in real RS where the hitbox is where you start the obstacle, but it makes more sense to have it be bigger for rs clicker.

CurtisGreen (approved):

PR #92, opened 2023-08-02
map zooming and right click menu change

EthanVieira:
* allow zooming on map with mouse scroll
* place item name in RCM in the middle of remaining space

### DIFF
--- a/src/clickable-object.js
+++ b/src/clickable-object.js
@@ -1,4 +1,5 @@
 import { CONSTANTS, FONTS, SCREEN } from "./constants/constants.js";
+import { getEstimatedPixelLength } from "./utilities.js";
 
 export class ClickableObject {
     // Basic info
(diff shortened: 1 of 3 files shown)


src/clickable-object.js
```diff
@@ -24,11 +25,21 @@ export class ClickableObject {
...
         let halfHeight = this.RCM_HALF_HEIGHTS[numActions - 1];
 
+        let actionTextSize = 0;
+
+        actions.forEach((action) => {
+            actionTextSize = Math.max(
+                actionTextSize,
+                getEstimatedPixelLength(action.text)
+            );
+        });
```
CurtisGreen:
```js
const actionTextWidth = actions
    .map((action) => getEstimatedPixelLength(action.text))
    .reduce((widest, current) => Math.max(widest, current), 0);
```

src/utilities.js
```diff
@@ -159,6 +159,33 @@ export function getItemText(amount) {
...
+                totalLength += 8;
+                break;
+            case str[i] == str[i].toUpperCase():
+                totalLength += 8;
+                break;
+            default:
+                totalLength += 6;
+                break;
+        }
+    }
```
CurtisGreen: Recently I've been liking the if return, if return... format over switch case in times like this where switch isn't actually used

```js
export const getEstimatedPixelWidth = (str) => 
    str.split("").reduce((sum, char) => {
        if (["i", "l"].includes(char)) return sum + 1;
        if (["m", "w"].includes(char)) return sum + 8;
        if (char == char.toUpperCase()) return sum + 8;
        return sum + 6;
    }, 0);
```

src/scenes/world-map.js
```diff
@@ -201,6 +202,26 @@ export class WorldMapScene extends Phaser.Scene {
         this.events.on("shutdown", () => {
             this.scale.resize(SCREEN.WIDTH, SCREEN.HEIGHT);
         });
+
+        container.on("wheel", function (pointer, deltaX, deltaY, deltaZ) {
```
CurtisGreen, line 206: Overall this is working but there are some weird edge cases at the edge of the map. I think it's good enough right now but worth keeping in mind

EthanVieira: Agreed. I struggled with some weird behavior for awhile and not sure what is the best solution for what we have, but this works good enough for me

CurtisGreen (approved):

PR #93, opened 2023-08-06
Draggable items

EthanVieira:
* Allow items to be draggable and swappable within the full inventory
* Would prefer a thorough test of this as there were many many issues when implementing.

### DIFF
--- a/src/items/equipment/equipment.js
+++ b/src/items/equipment/equipment.js
@@ -108,7 +108,7 @@ export default class Equipment extends Item {
         if (this.equipped) {
             console.log("Trying to unequip", this.name, this.index);
 
-            let wasAdded = this.scene.inventory.addToInventory(this, false);
+            let wasAdded = this.scene.inventory.addToInventory(this, true);
             if (wasAdded) {
                 this.scene.equipment.unequipItem(this.slot);
 
--- a/src/items/item.js
+++ b/src/items/item.js
@@ -57,18 +57,20 @@ export class Item extends ClickableObject {
             .image(x, y, spritePath)
             .setScale(this.scale)
             .setDepth(4)
-            .setInteractive()
+            .setInteractive({ draggable: true })
             .on("pointerdown", (pointer) => {
                 // Need to make sure left button isn't down to fix a bug where left clicking
                 // immediately after would trigger this twice and on the second time the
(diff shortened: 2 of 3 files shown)


CurtisGreen:
https://i.imgur.com/rZCDKkj.png

## Issues with counter
I'm not able to reproduce it consistently but sometimes gold (presumably any item with a counter) gets into a weird state:
- When you drag the gold the counter doesn't move with it
- If wouldn't let me move it up but it would let me move it to the side (see my cursor, i'm in the middle of a drag there)

I'll keep trying to find the reproduction case for a bit

## Issues with selecting
Dragging & selecting items don't play well together
- Click an item to select it
- Drag that item somewhere
- Click a different item to select it

Expectation: Either the drag removes the first select or the 2nd select tries to use the items on each other
Actual: Both items are selected. You can repeat this an infinite times to select all items in your inventry

EthanVieira:
> https://i.imgur.com/rZCDKkj.png
> 
> ## Issues with counter
> I'm not able to reproduce it consistently but sometimes gold (presumably any item with a counter) gets into a weird state:
> 
> * When you drag the gold the counter doesn't move with it
> * If wouldn't let me move it up but it would let me move it to the side (see my cursor, i'm in the middle of a drag there)
> 
> I'll keep trying to find the reproduction case for a bit
> 
> ## Issues with selecting
> Dragging & selecting items don't play well together
> 
> * Click an item to select it
> * Drag that item somewhere
> * Click a different item to select it
> 
> Expectation: Either the drag removes the first select or the 2nd select tries to use the items on each other Actual: Both items are selected. You can repeat this an infinite times to select all items in your inventry

Fixed the second bug when selecting.

For the first bug, I do see some small inconsistencies with where the item amount text is being rendered in relation to the item image. And sometimes when moving an item to the first column, the image will be rendered in the second column instead, but the text will be in the first column where it should be (and where it is logically in the inventory data). I tried to track that one and couldn't figure it out. It seemed like the values that were getting passed in to the phaser calls were correct, but when the animator is called it moves it to the incorrect spot sometimes... Might be able to fix it if we use containers to group the text and image, but when I tried doing that in August it seemed like a pain.

CurtisGreen (approved):

PR #95, opened 2024-01-23
two small changes

EthanVieira:
* burn food when using an already cooked version on the fire
* show the amount of an item in inventory when examining if the amount is >= 100k

### DIFF
--- a/src/ui/chat-window.js
+++ b/src/ui/chat-window.js
@@ -317,7 +317,7 @@ export class ChatScene extends Phaser.Scene {
     }
 
     // Show object info in chat window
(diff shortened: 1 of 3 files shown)


src/ui/chat-window.js
```diff
@@ -317,7 +317,7 @@ export class ChatScene extends Phaser.Scene {
     }
 
     // Show object info in chat window
-    showObjectInfo(isVisible, object = false, isShop = false) {
+    showObjectInfo(isVisible, object = undefined, isShop = false) {
```
CurtisGreen, line 320: Idk why I had it as false lol

CurtisGreen (approved):

PR #96, opened 2024-01-26
Runecrafting

EthanVieira:
* Added the Air Altar scene
* Started the Runecrafting skill
* I noticed a bug where selling all of an item messes up the inventory for any subsequent searches. Made a trello ticket for it
* Bonus: The serene track is S tier
* 
image


CurtisGreen (commented):
Should the talisman be added to the shop?

EthanVieira:
> Should the talisman be added to the shop?

I think it should be kept as a monster drop like it is in the real game

CurtisGreen (approved):

PR #97, opened 2024-01-31
Testing Updates

EthanVieira:
* added script for quickly running RsClicker on Windows/Chrome
* updated the load command to start the scene that is relevant to the test data
* fixed bug where searching the inventory when it has empty spots fails to complete


CurtisGreen (approved):
neat

PR #98, opened 2024-05-04
add sawmill scene to convert logs to planks

EthanVieira:
image


src/targets/enemies/goblin.js
```diff
@@ -1,5 +1,6 @@
 import { Enemy } from "./enemy.js";
 import NormalBones from "../../items/bones/normal-bones.js";
+import AirTalisman from "../../items/tools/talismans/air-talisman.js";
```
CurtisGreen, line 3: Was this supposed to be included previously?

src/targets/sawmill-operator.js
```diff
@@ -0,0 +1,88 @@
...
+        const chat = this.scene.scene.get(CONSTANTS.SCENES.CHAT);
+
+        // See if logs are selected
+        const selectedIndex = inv.curSelectedItemIndex;
+        const selectedItem = inv.inventory[selectedIndex];
+
+        if (
+            selectedIndex < 0 ||
+            !selectedItem ||
+            !this.validMaterialsToCostMap.hasOwnProperty(selectedItem.name)
```
CurtisGreen: any reason to use `hasOwnProperty` over bracket notation?

Ex: 
```js
!this.validMaterialsToCostMap[selectedItem.name]
```

CurtisGreen (commented):
Saw this error message when the level loaded

`this.targets[this.currentTargetIndex].updateProgress is not a function`

src/targets/enemies/goblin.js
EthanVieira: Yep lol

EthanVieira:
> Saw this error message when the level loaded
> 
> `this.targets[this.currentTargetIndex].updateProgress is not a function`

looks like an issue with autoclickers.. also affects other levels that require selecting something in inventory. I'll try to fix.

Thinking about it more.. maybe we don't want autoclickers to work when performing an action that uses a resource? It would auto-consume resources that the player may not want to be consumed. The alternative would be to have a on/off switch for them. Having it only work on targets that don't require resources would fit more into the theme of clanmates since clanmates can't actually help with most skilling. Thoughts?

CurtisGreen:
> Thinking about it more.. maybe we don't want autoclickers to work when performing an action that uses a resource? It would auto-consume resources that the player may not want to be consumed. The alternative would be to have a on/off switch for them. Having it only work on targets that don't require resources would fit more into the theme of clanmates since clanmates can't actually help with most skilling. Thoughts?

I agree, I think it makes sense to exclude things kinds of things

CurtisGreen (approved):

PR #99, opened 2024-05-05
add construction skill

EthanVieira:
* created the PoH scene and started the construction skill
    * the background/map screenshots are kinda low res because I took them from youtube videos.. need membership to get good screenshots
* created a trello task to fix quest completion for scenes with a single resource consuming target that makes different things for different quests
image


CurtisGreen (approved):
LGTM

PR #100, opened 2024-05-12
fix quest completion for levels with static targets that affect multiple different quests

EthanVieira:
100th PR hype


CurtisGreen (approved):
100th PR hype!

PR #103, opened 2025-02-22
Spell damage

EthanVieira:
* moved all assets outside of the src directory
* fixed casing for defenceBonus variables
* now using max hit and xp gain of specific spell when using magic.
* enemy.js and spellbook.js are the real files to review in this big diff


CurtisGreen (changes requested):
image

Something up with the tool images

EthanVieira:
> image
> 
> Something up with the tool images

lol resized them

CurtisGreen (approved):

PR #104, opened 2025-03-13
Combat styles

EthanVieira:
* Added dynamic combat style options in the combat style panel based on weapon type. These dictate both attack style and xp gained.
* TODO: need to calculate combat level and display it at the top of this panel.
image
image
image
image


CurtisGreen (approved):
Works perfectly nice job 😎

PR #106, opened 2025-07-21
add first prayer icons and ui functionality for turning them on/off

EthanVieira:
* prayer icons will only show in panel if you have the required level
image


src/scenes/load.js
```diff
@@ -1,5 +1,6 @@
 import { CONSTANTS } from "../constants/constants.js";
 import { AVAILABLE_INDEX, SPELL_MANIFEST } from "../constants/spell-manifest.js";
+import { PRAYER_MANIFEST } from "../constants/prayer-manifest.js";
 import { characterData } from "../cookie-io.js";
 import { itemManifest } from "../items/item-manifest.js";
 import { targetManifest } from "../targets/target-manifest.js";
```
CurtisGreen, line 6: Do you think all manifests should be styled like constants?

EthanVieira: I don't know what the convention usually is for those, but they are constant so I like styling them as constants

CurtisGreen (approved):

PR #107, opened 2025-07-23
add altar scene and implement prayer drain

EthanVieira:
* Move level up sfx to their own folder
* Add Lumbridge Altar scene for recharging prayer
* Implement prayer drain


src/scenes/world-map.js
```diff
@@ -34,6 +34,10 @@ export class WorldMapScene extends Phaser.Scene {
             .setDepth(2)
             .setInteractive()
             .on("pointerup", () => {
+                characterData
+                    .getScene(CONSTANTS.SCENES.DASHBOARD)
+                    .prayer.stopPrayerDrain();
```
CurtisGreen, line 39: I think prayer drain should probably be done when you enter world map rather than when you exit it. Unless I'm misreading this

EthanVieira: Whenever we start a level scene it seems like the dashboard scene and thus the prayer class is recreated but the original background prayer drain process continues even if the class was deleted (and no explicit destructors in JS so I can clean it up?). So this is a way of stopping the existing prayer drain before starting a new one.. There is almost certainly a better way to do this, but I figured we'd have more background processes in the future and I'd think of a way to handle them all when that is the case.

CurtisGreen (approved):

PR #109, opened 2025-08-09
add lumbridge thieving scene

EthanVieira:
image


CurtisGreen (approved):
This one looks so funny lol

I think you should have a way to add a hand or something that makes it look like he's pickpocketing

